PR #9, opened 2018-02-24
Merged dbm further commits through a66abb2  on 26-FEB-2018

easye:
This pull merges @dbmcclain's further commits on 24-FEB-2018 (my Saturday afternoon).

Travis CI has been reduced to invoking only `sbcl`, `ccl`, and `abcl` as we have a limited amount of time for our builds, and these are the most useful implementations for diagnostics at the moment.

PROVE and LISP-UNIT tests now should be comfortably living together in the source tree to the point that developers who wish to have nothing to do with PROVE should not need the `(ql:quickload :prove)` hack *unless* they wish to run the tests in `emotiq-prove` or `cosi-prove`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,11 @@ env:
   matrix:
     - LISP=sbcl-bin
     - LISP=ccl-bin
-    - LISP=abcl-bin
-    - LISP=ecl
-    - LISP=clisp
-    - LISP=clasp
+# Uncomment to enable the build farm resource to be spent on implementations that might actually produce usefull results    
+#    - LISP=abcl-bin
+#    - LISP=ecl
+#    - LISP=clisp
+#    - LISP=clasp
 
 install:
   - curl -L https://raw.githubusercontent.com/snmsts/roswell/release/scripts/install-for-ci.sh | sh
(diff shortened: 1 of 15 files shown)


dbmcclain (commented):
test-package.lisp was written to succinctly demonstrate a possible ANSI violation in LW. The issue is that importing the same underlying symbol via two different paths produces a compile error in CCL but is accepted by LW. Shannon thought LW was violating ANSI, but there is room for disagreement.

At any rate, there is no reason to keep test-package.lisp alive except for historical curiosity.

I looked at the diffs shown for the edwards.lisp and they look valid but dated. Meaning they were valid for some earlier commit, but code has progressed since then. Same for ecc-package.lisp, and cosi-keying.lisp

easye:
@dbmcclain 

test-package.lisp was my starting to get the LISP-UNIT machinery working with COSI.  It showed me that LISP-UNIT needs to be invoked with appropriate handlers if the underlying code itself has compatibility problems (I believe that the DEFPACKAGE forms in test-package.lisp *only* compile under LLW; they definitely don't compile under `ccl`).

The dated inconsistencies edwards.lisp ,ecc-package.lisp, and cosi-keying.lisp are because you updated the `dbm` branch with further changes since this pull request was originated.  I will merge your additional changes to this branch for your further review when I get a chance.

easye:
@dbmcclain I have added your further commits on 25-FEB-2018 to this pull request.  This should remove the "dated" problems you noticed.  Please let me know if this is not the case.

Your changes to `package.lisp` from my hacks around symbols  unfortunately fail on the open ANSI implementations (sbcl/ccl) that we are testing coverage upon (cf. the Travis-CI results for `ccl` at https://travis-ci.org/emotiq/emotiq/jobs/345854035 See https://github.com/emotiq/emotiq/pull/9/commits/b9f68d4bc87d862ddb9d64d3d1ee822b2af2f834 for my comments on the current state of `package.lisp`

vlebedev (approved):

PR #11, opened 2018-02-26
copy blockchain.lisp with package from research, add to emotiq asd

markhdavid:


dbmcclain:
Again, don't know what I'm doing here... Looks okay to me?

easye:
@dbmcclain I don't think we have established what a review means yet, so it is mostly an advisory mechanism so we know something is going into the `dev` stream.  

Since others expressed interest in @markhdavid's strawchain proposal, asking people to review the code at least bumps it into their visibility.

easye (approved):
Please delete the corresponding code from emotiq/research so we don't get confused over the current version.

PR #13, opened 2018-03-02
Cosi tweaks

markhdavid:
This allows CCL to build COSI.


markhdavid:
should have been off dev

PR #20, opened 2018-03-04
Fix reader macro issues on SBCL

markhdavid:
The #+SBCL reader-conditioinalized code should help with SBCL 2.2
issues. To the extent that it helps, we thank Christophe's informative
blog post here
http://christophe.rhodes.io/notes/blog/posts/2014/naive_vs_proper_code-walking/
Note that he gave numerous caveats and warnings, which we are for now
disregarding, so we are responsible if this of limited value, as he
explains in his blog at length.

While diving in I rubbed my eyes a lot and figured out what the heck the example
was meant to be.  So I rewrote it for clarity.  Hopefully, it's clearer.

### DIFF
--- a/src/useful-macros/basic-useful.lisp
+++ b/src/useful-macros/basic-useful.lisp
@@ -105,14 +105,25 @@ THE SOFTWARE.
 
 ;; ----------------------------------------------
 
+#+sbcl
+(if (string-lessp (lisp-implementation-version) "1.2.2")
+    (pushnew :safe-sbcl *features*)
+    (setq *features* (remove :safe-sbcl *features*)))
+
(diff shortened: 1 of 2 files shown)


dbmcclain (commented):
I'll take your word for it, for now... I'd have to dig into SBCL SB-IMPL::COMMA to understand the issues. Also, that's an internal symbol, and so probably not a great idea to leave this in the sources for production use...

markhdavid:
From what David said above and in messaging on Twist, he approves this, though he seems to have missed ticking off the [x] approve button, since it's easy to miss, and he's new to Github and this workflow. So I'm going to proceed in the spirit of his intent and put this through now.

PR #21, opened 2018-03-04
define & use defconstant+ for non-eql-comparable values (SBCL needs)

markhdavid:
SBCL stops the show if you defconstant with a value that can't be compared using EQL, that is, on the second and subsequent evaluations it throws an error about it. So I defined defconstant+. If this is a subsequent evaluation, it just sets the value of the constant to the identical (old) value, never comparing at all.


markhdavid:
Closing to rebase this on top of Fix FLATTEN for SBCL.

PR #24, opened 2018-03-06
Changes to modular-arithmetic

dbmcclain:


markhdavid (approved):
One change is just commenting out the use of a module.
The other is slew of changes to a new file under active development not used anywhere yet. So approved.

PR #26, opened 2018-03-06
Fix 1st arg to MAP, must be sequence type, not function. SBCL NOT amused

markhdavid:


markhdavid:
going to try to redo on top of defconstant+ fixes

svspire:
This is one of those bugs that falls into the category of "How did this ever work before?" Nice job tracking it down.

markhdavid:
Thanks. Well, it was an unused function. Well, here's an example of even unused code having a cost.

PR #27, opened 2018-03-06
define and use defconstant+ for non-eql-comparable values (SBCL needs this)

markhdavid:


markhdavid:
Ugh, sorry, git(hub) confusion. Closing!

PR #28, opened 2018-03-06
Fix 1st arg to MAP, must be sequence type, not function. SBCL NOT amused

markhdavid:

### DIFF
--- a/src/useful-macros/useful-macros.lisp
+++ b/src/useful-macros/useful-macros.lisp
@@ -1841,7 +1841,7 @@ This is C++ style enumerations."
   (values (read-from-string (apply #'mkstr args))))
 
 (defun explode (sym)
-  (map #'list (compose #'intern-symbol #'string) (symbol-name sym)))
+  (map 'list (compose #'intern-symbol #'string) (symbol-name sym)))
 
 ;; -------------------------------------------------------
 ;;


markhdavid:
So I did finally push defconstant+ to dev get that part of SBCL going better. Then I had had this branch out there but prior to that being there. So I at first had a few false starts before I realized I should do a force push. This is OK for a private, short-lived bug-fix branch, but I'd forgotten.  Anyhow, now that's done. So these are notes mostly for @easye @vlebedev @emtq : we should probably discuss conventions around fully private vs. team branches. You want people pushing branches to github, and there should just be assumptions around "private" branches. So a push to gihub should not mean they have to give up the ability to squash and rebase and curate their commits ("change history"). Of course, for "dev" and "master" basically they must "never" do so, and then there will be team branches that are in between cases.

svspire:
Here's a slightly simpler definition:

    (defmacro defconstant+ (name value &optional doc)
      `(defconstant ,name 
         (if (boundp ',name) ,name ,value)
         ,@(if doc (list doc))))

markhdavid:
Thank you. Good observation @svspire, although it's about a previous commit/pull request. @svspire would you like to approve this PR? (For the #'list => 'list change?)

easye (approved):
With this diff `useful-macros` still doesn't compile under `sbcl` due to the need to apply the multi-processing fixes, but the compilation gets further before bombing out in `uuid.lisp`

I will get the multi-processing patch in after this later today.

PR #30, opened 2018-03-07
Addition of Okeanos sources, growth in RandHound

dbmcclain:


markhdavid:
This was done wrong by my and our collective mistake. We should have created a branch. Now commits have just been accumulating. There's no point to using this. 

I do want to get rid of the backlog of dbm branch now, but I may as well do it via github's already provided suggested automatic merges from dbm onto dev, rather than this 3-day-old stale PR.

Closing.

PR #33, opened 2018-03-08
Fix open ANSI compilations across ccl sbcl (partial)

easye:
In contrast to `ccl`, `sbcl` has never compiled cleanly, but the failure @mhd observed in `useful-macros` actually occurs in `uuid.lisp`. Something is futzed with sbcl/SLIME/ASDF reporting at the moment, so it isn't clear this isn't the same error.

In any event, with this pull request I have added my current state of porting to  `sbcl` which compiles `useful-macros` completely, but then currently gets bogged down in `core-crypto`.

For `ccl`, I had to conditionalize a type specifier in f7f92f0 @dbm @svspire


markhdavid (approved):
approved

markhdavid (changes requested):
Actually, @easye, I had taken your word for it before and not actually tried out the branch, but just now I checked out the branch, and I could still not get past SBCL reader macro bug that had been fixed. It says COMPILE-FILE-ERROR while
compiling #<CL-SOURCE-FILE "useful-macros" "ppcre-reader">
If I'm missing something, can you please provide more detail. Thank you.

EDIT/UPDATE: ugh, figured it out: I had put a bad repo "to the side", and it was getting picked up due to my liberal ASDF configuration without my noticing it, so, kind of a form of pilot error. Anyhow, the above was not really happening with this branch, but with a random branch I was fooled into believing was this branch. Sorry. And so I've changed this back to being approved. Please merge into dev.

markhdavid (approved):

PR #36, opened 2018-03-12
SBCL - last tango

markhdavid:
Some last non-LispWorks changes fixing some net utils, fix 2 defconstants.
The non-LispWorks changes are minimal changes, not wanting to take precious time to review.
But these were done and ready at the tail end of when we were still trying to get SBCL to work,
so may as well get them in before putting SBCL in the freezer for a few months.
FWIW on SBCL system "cosi" actually builds to completion in this branch, assuming you can convince ASDF to NOT signal an error due to the existence of non-style warnings (which I know how to do, but it's involved; see research branch).

### DIFF
--- a/src/Cosi/base58.lisp
+++ b/src/Cosi/base58.lisp
@@ -35,7 +35,7 @@ THE SOFTWARE.
 
 ;; from https://bitcointalk.org/index.php?topic=1026.0
 
-(defconstant +alphabet+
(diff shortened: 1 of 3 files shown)


markhdavid:
Squashing, back soon... and ... I guess I do not have 'PR reopen privileges', or else I would reopen. So I'll just create a new PR. OK.

PR #38, opened 2018-03-13
last sbcl changes, part 2

markhdavid:
Mark E pointed out that the last set of commits still left actual compiler errors. Well, a quick glance revealed actual real errors in code that should not be there in any Lisp, and now I think SBCL compiles without errors.


dbmcclain (commented):
looks okay to me

easye (approved):

PR #39, opened 2018-03-14
Successful test runs show up as green in the CI

easye:


markhdavid:
I think it should return failure (non-zero) as soon in the case where the test fails, i.e., in the case where (asdf:test-system ...) returns false (nil) rather than falling through and doing who knows what. (We only figure we know because we spent the last several weeks experiencing it mysteriously returning code 1 when we let it fall through, I guess.) Also, I suggest documenting the return value 0 for success and where we return the value for failure have it be a non-zero return value that's documented and consistent, either with a defconstant or at least a comment.

markhdavid (changes requested):
I think it should return failure (non-zero) as soon in the case where the test fails, i.e., in the case where (asdf:test-system ...) returns false (nil) rather than falling through and doing who knows what. (We only figure we know because we spent the last several weeks experiencing it mysteriously returning code 1 when we let it fall through, I guess.) Also, I suggest documenting the return value 0 for success and where we return the value for failure have it be a non-zero return value that's documented and consistent, either with a defconstant or at least a comment.

easye:
@markhdavid Please review the requested changes in https://github.com/emotiq/emotiq/pull/39/commits/b5a07ff1893155fae17465f47a41d33084bb5932

.travis.yml
```diff
@@ -31,16 +30,19 @@ install:
...
-               (uiop:quit 0))'
-  # Speculatively hook into Coveralls <https://coveralls.io/>
-  # Coveralls invocation (currently unused)
-  #                 
-  # - ros -s prove -s cl-coveralls
-  #       -e '(or (coveralls:with-coveralls (:exclude (list "t"))
-  #                 (prove:run :emotiq-prove))
-  #               (uiop:quit -1))'
+  - ros -e "(loop
+              :for i :upfrom 0
```
markhdavid, line 52: how about upfrom 1

.travis.yml
```diff
@@ -31,16 +30,19 @@ install:
...
+                  (system dependencies) 
+              :in 
+                  '((:emotiq :emotiq-test)
+                    (:cosi   :cosi-test))
+              :doing 
+                 (ql:quickload dependencies)
+              :unless 
+                 (asdf:test-system system)
+              :return 
+                 (uiop:quit (ash -1 i))
```
markhdavid, line 63: how about quit with i

.travis.yml
```diff
@@ -31,16 +30,19 @@ install:
...
+              :finally 
+                 (uiop:quit 0))"
```
markhdavid, line 65: ;; success

markhdavid (changes requested):
I think I found a way to get a bit simpler, no need to bit shift, and add a comment on exit status, please; see my github  comments inline

.travis.yml
easye: The `ash` forms gives different exit status depending on which test has failed.

.travis.yml
easye: Because only FORTRAN programmers know what `option base` does.

And `(ash -1 1)` is `-2` whereas `(ash -1 0)` is `-1`

markhdavid: I don't get it. Do I have to know a lot about Fortran to understand that? Oh well, I suggested it to go with a change below, but I'll "fix" it below...

.travis.yml
markhdavid: OK, yes, agree it can be useful to indicate which test has failed - good idea. But I don't think it's needed to shift bits for each test number. With not too many tests, it could risk getting into bignums (in Lisp) and into 32 and 64+-bit integers in Unix, which may not work well for interprocess communication. Also, don't see use for negative numbers Suggest just returning test index with 1 added to it to make it be non-zero, so can just do (uiop:quit (1+ i)) in the failure case.

.travis.yml
markhdavid: We could some day do a fancier status code with both a run number (the 'i') and a type-of-problem report. I wrote a prototype Lisp function for this in a research branch for this. Check it out: https://github.com/emotiq/research/blob/status-code-ci/status-code-ci.lisp
But I'd suggest deferring on that for now and go with the simpler approach above.

easye: @markhdavid Total bikeshed on the return value, but nice code!  

Standard POSIX return value <https://en.wikipedia.org/wiki/Exit_status#POSIX> but I wouldn't rely on more than 8 bits of status.

markhdavid (approved):
OK. As discussed, we can fine-tune later.

PR #48, opened 2018-03-26
Lispworks working under CI

easye:


.travis.yml
```diff
@@ -17,7 +17,7 @@ env:
     - LISP=sbcl-bin
 # Uncomment to enable the build farm resource to be spent on
 # implementations that might actually produce useful results.
-    - LISP=abcl-bin
```
markhdavid, line 14: Also commented out ABCL build

markhdavid (approved):
As long as you intended to remove the apt-get install line in one of the changed yaml files, this is approved. 

I also see you commented out ABCL build, which is good move for now I think.

PR #49, opened 2018-03-28
Gossip

svspire:
New version of gossip protocol. Using actors now. (Old version can still be loaded by using simple-gossip.asd.)
No-reply-necessary messages are all working.


easye (approved):
It would be nice to have the forms outlined in readme.txt put into a test fixture that I could exercise, but otherwise I can get it to run.

PR #50, opened 2018-03-28
Build and deliver PBC shared objects under linux/macos

easye:
Merge pt-dbm-mte


guitarvydas (approved):
Approving this so MarkE can merge to dev.

PR #52, opened 2018-03-28
Basic script evaluation with authentication for spend transactions

markhdavid:
Basic script evaluation with authentication for spend transactions


easye (approved):
It would be good to start splitting the script evaluation into a separate compilation unit, so Paul can start working on that section of things independently.

PR #53, opened 2018-03-29
Instruct the CI to only run LispWorks as agreed

easye:
Developers may renable test coverage for tests for a given
implementation by uncommenting the necessary lines to this script in
their development branch.


markhdavid (approved):
This is of course in line with the company's decision to go with LispWorks-only for the short term, to get our first release established, etc., and avoiding time-consuming projects needed for the other platforms (which came along unfortunately)

PR #56, opened 2018-04-03
Gossip

svspire:
Handles replies.
Tests present and working.

### DIFF
--- a/src/gossip/gossip-tests.asd
+++ b/src/gossip/gossip-tests.asd
@@ -1,0 +1,19 @@
+;;; -*- Mode: LISP; Syntax: COMMON-LISP -*-
+
+(defsystem "gossip-test"
+  :version "0.0.1"
+  :description "Gossip Tests"
+  :author "Copyright (c) 2018 Emotiq AG"
+  :license "MIT (see LICENSE.txt)"
+  :depends-on (gossip lisp-unit)
+  :serial t
+  :perform (test-op (o s)
+             (symbol-call :lisp-unit :run-tests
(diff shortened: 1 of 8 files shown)


easye:
The tests don't execute
```(asdf:test-system :gossip-tests)```
fails with
```Package GOSSIP-TEST not found```

easye:
Additionally, it please add an `:in-order-to` clause in the system that is exercised in the tests as in

https://github.com/emotiq/emotiq/blob/dev/src/emotiq.asd#L11

easye (changes requested):
Please fix it so that the tests can be run.

svspire:
Try pulling my latest push instead. I think I fixed the typo.
As for :in-order-to, I have no idea what that clause or :perform is for, and the ASDF docs are impenetrable to me. So please feel free to correct those yourself.

easye:
@svspire 

The `:in-order-to` clause at https://github.com/emotiq/emotiq/blob/dev/src/emotiq.asd#L11 tells that in order to perform the tests on the `emotiq` system to run test-op on the `emotiq-test` system.

You already have a `:perform` clause in https://github.com/emotiq/emotiq/blob/gossip/src/gossip/gossip-tests.asd#L10, which is how you are getting 

```(asdf:test-system :gossip-test)``` 

to run.

If you were to add the `:in-order-to` clause in `gossip.asd`, you would be able to invoke these tests via

```(asdf:test-system :gossip```

easye (approved):
At least the tests can be invoked now, if not quite in the way I requested.

PR #57, opened 2018-04-03
Dbm reentrant pbc

dbmcclain:
Major simplifications for everyone. PBC is now effectively reentrant. No more need for WITH-CRYPTO, ASK-CRYPTO. Simplifies use of PBC crypto everywhere.


easye (approved):
N.b. Current ASDF machinery will not rebuild crypto-pairings library when underlying C code changes.  How do we wish to coordinate?

PR #58, opened 2018-04-04
tx test code isolated, less globals, coinbase, genesis, staking progress

markhdavid:
Blockchain test-demo code now in a new separate lisp-unit test file.
There's now at least one unit test for transactions, so that's a
start.  The globals have been lessened a bit while keeping everything
working, including the demo, a must.  There is some code for staking
but not all functioning, but some ideas to work on and discuss
further.  There is a more sensible Genesis block, and there are units
and subunits of the coin that make some sense now, though not the
final word. Lots of cleanups.

---
** can do the usual for unit tests: (asdf:test-system "emotiq") **
** and then can run the demo like so: (emotiq-test:test-blockchain) **


easye (approved):
Looks good.

PR #60, opened 2018-04-04
Single chain block structure in CLOS

easye:


markhdavid (approved):
I did minimal testing. I could still load my old transaction code and tests same as before. Approved.

easye:
@markhdavid 

Example of a messed up pull request, which cannot be merged into `dev` because the state of `dev` has evolved into something beyond the capability of the automatic merge tools.  

The use of the "web editor" options would not allow one to test the source changes before pushing to `dev`

One cannot follow the "command line instructions" because they involve pushing to `dev` which isn't allowed (I tried it out, and got error messages).

This is a trivial change: I can easily redo the pull request.  But we should figure out what the right steps "should have been" now, until we have really complicated merges.

PR #64, opened 2018-04-05
Pt initc

guitarvydas:
Two changes to do DLL (CCFFI) initialization at runtime (need-pairing) instead of at compile-time.  On Mac, the symptom was an error on dev-start that looked like a failure to load libLispIntf.dylib (the error wasn't *during* the load, the error was caused by the code simply *not loading* the dylib).

PR #65, opened 2018-04-05
Pt initc rebase

markhdavid:
This was spun off of PR #64 which had conflicts and needed rebasing with dev. That had been on branch pt-initc. This is on branch pt-initc-rebase which is identical but rebased with dev HEAD at commit 3c8d1a3 and pt-initc HEAD at commit 969fdb3.  Last 4 commit SHAs of pt-initc-rebase are new.


markhdavid:
I just minimally tested it, after creating this through some git manipulations. Paul should now do a sanity check.

PR #67, opened 2018-04-06
Pt6 deliver

guitarvydas:
I just did 'git rebase dev' with this branch pt6-deliver (already up-to-date).

This is the full set of changes to deliver on linux and on recent Mac's.  It still doesn't work on a 2010 Airbook.


src/Crypto/pbc-cffi.lisp
```diff
@@ -121,17 +121,37 @@ THE SOFTWARE.
...
+  "loads the DLLs (.so and .dylib) at runtime, from the current directory"
+  (cffi:define-foreign-library
+   libpbc
+   (:darwin "libLispPBCIntf.dylib")
+   (:linux  "./libLispPBCIntf.so")
+   (t (:default "libLispPBCIntf"))))
+
+(defun load-dlls()
+  "load the dev or production dlls at runtime"
+  (if (emotiq:production-p)
```
markhdavid, line 151: This is a problem: we ought to be able to build "cosi-bls" system standalone.  That system does not currently have a dependency on emotiq, and no emotiq package gets created, so if you simply start up fresh and do (asdf:make "cosi-bls"), you get a package error for emotiq package unknown.  The CI is not helping us in that it masks defects in our system definitions: in this case it quickloads emotiq first, so the package happens to be sitting around. For now, rejecting this PR due to this.

markhdavid (changes requested):
Rejecting because even though it builds if you happen to quickload in the right order, it has code in a package (emotiq) that's not declared anywhere in the cosi-bls system or its dependencies. Let's discuss tomorrow. I guess it's the new production-p flag - understandably a new global concern. We will find a way.

src/Crypto/pbc-cffi.lisp
```diff
@@ -121,17 +121,37 @@ THE SOFTWARE.
...
-
-(cffi:use-foreign-library libpbc)
+;; The Lisp runtime load might help us do this differently, but explicit
+;; initialization is much easier to understand
+
+(defun load-dev-dlls ()
+  "loads the DLLs (.so and .dylib) at runtime, from pre-specified directories"
+  (cffi:define-foreign-library
+   libpbc
+   (:darwin #.(concatenate
```
easye, line 131: @guitarvydas Better to use `cl:merge-pathnames` than to convert out to namestring.

.gitignore
```diff
@@ -1,3 +1,4 @@
 *.fasl
 *.64xfasl
+*.64ufasl
 *~
```
easye, line 4: @guitarvydas Add "var/*"?

easye (approved):

src/Crypto/pbc-cffi.lisp
easye: I addressed this problem with 

https://github.com/emotiq/emotiq/commit/9177c8cd352c634884500238798f0d4cbcc4ac87

@guitarvydas This is the commit I referenced in the Twist.

markhdavid:
Closing this in favor of new PR #75

PR #73, opened 2018-04-06
Ci nuke fasls between tests --  Remove the cached fasl artifacts between test-system invocations

easye:
Isolates inter ASDF system dependencies to reveal incomplete
dependency declarations.

Developers are advised to add their ASDF definitions to the LOOP form
explicitly if they wish such coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,8 @@ script:
                     (:cosi-bls :cosi-bls-test)
                     (:crypto-pairings :crypto-pairings/t))
               :doing 
+              (progn
+                 (uiop:run-program "rm -rf ~/.cache/")
                  (ql:quickload dependencies)
               :unless 
                  (asdf:test-system system)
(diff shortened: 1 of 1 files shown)


markhdavid (changes requested):
- I called for something like this to be done, but I meant it as an adjunct to another approach, which I went ahead and prototyped in PR #78 . I think it's good enough, although perhaps too verbose code-wise. Hopefully Mark E can slim it down.
- My idea is that each system is built in a fresh Lisp. So each system proves it's got all dependencies met each run. (Then removing binaries on top of that helps prove it can also generate binaries from scratch in a fresh Lisp without the benefit of the side effects of having loaded the previous system.)
- In terms of removing binaries, it seems the right way to do it is via the ':force t' or ':force :all' args to the ASDF compiler operations, although we seem to be one step removed from ASDF due to Quicklisp, so I could see that as problematic in the short run.
- Even if doing the 'rm -rf' thing, it should only do ```~/.cache/common-lisp/```, not all of ```~/.cache/```, just to be cleaner and more conservative (```~/.cache``` being a general cache resource).
- Therefore, rejecting based on not matching the stated goal. I think this should be done after the dust has settled on PR #78, as a supplement to that only.

PR #75, opened 2018-04-07
Finishing touches for production (executable) builds.

markhdavid:
… [also dumps memory usage]

forgot to add startup.lisp

initialize dll's explicitly (so far, good only for dev)

wip: working for dev on linux

some vars missing

beginning to work on linux again

small changes for mac

removed debug statements

Finishing touches for production (executable) builds.

deliv-linux.bash, deliv-macos.bash - add switch to disable loading a
  personal init file in scripts

deliver.lisp - change Lisp start function to emotiq:start

crypto-pairings.asd - add dependency on emotiq/delivery

emotiq-test.asd - just bump version number

emitiq.asd - define new systems emotiq/delivery, emotiq/startup and
  bump version of emotiq system

package.lisp - minor changes in exports

production.lisp - new file, super simple, just defines *production*
  and production-p function

PR #76, opened 2018-04-07
Pt10 deliver

guitarvydas:
This appears to work for all 4 cases (linux dev, linux prod, mac dev, mac prod) and incorporates what was discussed this afternoon (production-p in a separate file, emotiq/delivery defsystem)

The word "production" might be confusing (poorly chosen).  For the purposes of this code, "production" means one thing and will never mean anything else.  There are two cases covered - developers should call "main" and the binary calls "start".  Period. It will never be meaningful to call "start" from a lisp development system (although, it might be possible to call it).  If the need to change the meaning arises, it will probably be bettr to change the name of the variable (e.g. to *binary* or *delivery*).

In this version, the binary calls "start".  It is trivial to hack the code for "start" to set *production* based on the command line.

FTR: pt7-deliver became "corrupted" when I tried to rebase it with dev.  9b-deliver is a pre-corruption save of pt7-deliver.  If this code, "pt10-deliver", makes it into dev, then we can delete all of that other stuff.


markhdavid:
Closing previous PR #75 (https://github.com/emotiq/emotiq/pull/75) in favor of this (#76 ).

markhdavid (approved):
Successfully build for macOS.

- to build:

  cd ~/repos/emotiq/etc; bash production.bash

- to run after build:

  cd ~/repos/emotiq/var/local/production-macos; bash emotiq.bash

markhdavid (changes requested):
Oh, I just noticed this PR says there can be no automatic merg with dev. Have to figure that out. Pulling back the approval until that can be resolved.

Message is:

>Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.

markhdavid (approved):

markhdavid (changes requested):
I have created a branch against dev by hand dev-hand-merge-pt10-deliver, and I'll create a superseding PR with that. So this will be closed I'm afraid.

PR #80, opened 2018-04-09
Dbm 0409b

dbmcclain:
work from last week


guitarvydas (changes requested):
You need to assign a reviewer for each pull request, which you forgot to do.  A pull request is not finished until it is reviewed and accepted.

See this new doc (the rest of my comments are related to the development/production split):

https://github.com/emotiq/emotiq/blob/binary-doc/etc/README

guitarvydas:
dfsfd

src/Crypto/pbc-cffi.lisp
```diff
@@ -155,7 +155,7 @@ THE SOFTWARE.
 
 (defun load-dlls()
   "load the dev or production dlls at runtime"
-  (if (emotiq:production-p)
+  (if nil ;; (emotiq:production-p)
```
guitarvydas: Production-p is defined in system emotiq/delivery.  I think that this system has not been loaded and, thus, caused problems

src/Crypto/pbc-cffi.lisp
```diff
@@ -153,11 +153,22 @@ THE SOFTWARE.
    (:linux  "./libLispPBCIntf.so")
    (t (:default "libLispPBCIntf"))))
 
+(defun load-ral-dlls ()
```
guitarvydas, line 156: "load-ral-dlls" should be removed.  We are trying to put pbc and gmp into local directories and not rely on the default build procedure (the default installs the libraries in /usr/local/, which is something we want to avoid. Better documentation (by me) is required.

src/Crypto/pbc-cffi.lisp
```diff
@@ -153,11 +153,22 @@ THE SOFTWARE.
...
+  (cffi:define-foreign-library
+   libpbc
+   (:darwin "/usr/local/libLispPBCIntf.dylib")
+   (:linux  "/usr/local/libLispPBCIntf.so")
+   (t (:default "libLispPBCIntf"))))
+
 (defun load-dlls()
   "load the dev or production dlls at runtime"
-  (if nil ;; (emotiq:production-p)
+  #-:COM.RAL
```
guitarvydas: The call to (emotiq:production-p) happens at runtime to decide whether we are in development mode or we are building a stand-alone binary as "production".  In development mode, we run a script which builds PBC and GMP dylibs and puts them in emotiq/var/local/lib.  The script needs to run only once.  It runs 'config', 'make' and 'make install' with switches that put the results into emotiq/var/loca/include and emotiq/var/local/lib.  As long as one doesn't 'rm -rf emotiq/var', the development environment remains.  When we want to build binaries, we wipe emotq/var and rebuild everything using another script.  This script puts everything needed (the binary emotiq and 3 dylibs - libLispIntf, libpbc.dylib and libgmpl.dylib and a script emotiq.bash - into a single directory which can be zipped and sent to users.

src/Crypto/pbc-cffi.lisp
```diff
@@ -153,11 +153,22 @@ THE SOFTWARE.
...
+  (if (emotiq:production-p)
       (load-production-dlls)
     (load-dev-dlls))
+  #-:COM.RAL
```
guitarvydas: remove this work-around

src/Crypto/pbc-cffi.lisp
```diff
@@ -157,8 +157,8 @@ THE SOFTWARE.
   "loads the DLLs (.so and .dylib) at runtime, from the current directory"
   (cffi:define-foreign-library
    libpbc
-   (:darwin "/usr/local/libLispPBCIntf.dylib")
-   (:linux  "/usr/local/libLispPBCIntf.so")
+   (:darwin "/usr/local/lib/libLispPBCIntf.dylib")
```
guitarvydas: Emotiq developers should be using emotq/var/local/lib/{libLispPBCIntf.dylib, libpbc.dylib, libgmp.dylib} and not be using /usr/local/*.

guitarvydas (changes requested):

PR #85, opened 2018-04-12
New protocols

svspire:
Lots of new code. :gossip vs. :neighborcast protocol. Network mechanisms are present.


markhdavid (approved):
Looks good.

PR #86, opened 2018-04-12
xxx ignore bad Ex-PR (basic UTXO ...)

markhdavid:
…e, remove obsolete code


markhdavid:
bad pull request on a branch for change that was then replaced after a forced merge

PR #88, opened 2018-04-13
Dbm 0412 transactions

dbmcclain:
transactions for Cosi, crypto proofs, validation, etc


markhdavid (approved):
looks good. needs rebasing. can be auto-rebased though (I checked.)

PR #95, opened 2018-04-17
Blocks mhd

markhdavid:


easye (changes requested):
@markhdavid Need to remove inadvertent check-in of emacs backup file  `emotiq/src/tests/#blockchain-test.lisp#`

markhdavid:
Review welcome, but do not merge please.  I may have a replacement or modifications soon.

markhdavid:
Good catch @easye 
I removed the offending file. Also fixed something in unit tests. (Forgotten commit, actually.)
Tested with: (asdf:test-system :emotiq-test) then (emotiq-test:test-blockchain)

markhdavid:
Mark, let me close this while David (et al) think about the version in Cosi-BLS, which is different and evolved from this in various ways. Thanks for the review.

PR #98, opened 2018-04-18
Add a block class to Cosi-BLS, package cosi/proofs.

markhdavid:


src/Cosi-BLS/block.lisp
```diff
@@ -0,0 +1,207 @@
...
+               (when (= int-so-far 0)
+                 (loop-finish)))
+        finally (return (nreverse result))))
+
+;; General, so move elsewhere, or merge/replace with something already
+;; in DBM's tool box! -mhd, 4/18/18
+
+
+
+(defun serialize-block-octets (block)
```
easye, line 133: It might be simpler to use the `lisp-object-encoder` (q.v. <https://github.com/emotiq/emotiq/blob/dev/src/lisp-object-encoder/lisp-object-encoder.lisp#L166>) code to round-trip a CLOS description to/from a stream of bytes.  The caveat with that approach is that you then probably not want a in-memory reference to the previous block like you have in the `prev-block` pointer.

src/Cosi-BLS/block.lisp
```diff
@@ -0,0 +1,207 @@
...
+   ;; Block-transactions is generally what's considered the main
+   ;; contents of a block whereas the rest of the above comprises
+   ;; what's known as the 'block header' information.
+   (transactions
+    :accessor transactions
+    :documentation "Sequence of transactions")
+
+   ;; Caller can set and maintain these slots. These could be set to
+   ;; contain a treap or possibly for now just a simple list.
+   (validator-keys-joining
```
easye, line 49: I thought that we had agreed to simply contain a list of all validator keys in the current epoch, rather than use the approach of noting the deltas of keys joining/keys leaving.

src/Cosi-BLS/block.lisp
```diff
@@ -0,0 +1,207 @@
...
+    :accessor validator-keys-joining
+    :documentation "The validators nodes joining in this epoch.")
+   (validator-keys-leaving
+    :accessor validator-keys-leaving
+    :documentation "The validator nodes leaving in this epoch."))
+
+
```
easye, line 56: We are missing (at least)

1)  A field recording the validation leader for the epoch

2) Information for recording the validator witness signatures which would consist (as I understand it) of:
   a)  a bitmap representation of all validators that have signed this block
   b)  the corresponding cryptographic proof of all co-sign witness

src/Cosi-BLS/block.lisp
```diff
@@ -0,0 +1,207 @@
...
+      )
+
+   (prev-block
+    :accessor prev-block
+    :documentation "Previous block (nil for genesis block).")
+   (prev-block-hash
+    :accessor prev-block-hash
+    :documentation "Hash of previous block (nil for genesis block).")
+
+   (merkle-root-hash
```
easye, line 31: As I understand it, if we use an authenticated treap structure to contain the transactions, such a hash is provided by querying the root object of that treap, so this field would be redundant.

src/Cosi-BLS/block.lisp
```diff
@@ -0,0 +1,207 @@
+;;;; block.lisp
+
+(in-package :cosi/proofs)
```
easye, line 3: It would perhaps be a better abstraction not to have our block structure inside the `cosi-bls` ASDF definition, but rather "up an an abstraction level" by moving such code up a directory, and within its own ASDF definition.  

The reasoning here, is that `cosi-bls` has been presented by David as containing the service api primitives that should be used to construct the basic chain we are after.  This would also imply that `transactions.lisp` should also be moved into a "higher abstraction" level, so would require coordination with the code of @dbmcclain

easye (approved):
None of my comments on @markhdavid's  current approach should actually block merging *something* to `dev` so we have a strawman to kick around, so I approve this merge.

src/Cosi-BLS/block.lisp
guitarvydas: I believe that the "previous" block hash is there to quickly "prove" that the chain (the blocks leading up to this current one) hasn't been tampered with. An ADS tree is a "super-class" of Merkle (ADS generalizes the Merkle strategy to other kinds of data structures).  A treap is just another kind of tree (https://en.wikipedia.org/wiki/Treap).

PR #102, opened 2018-04-18
implement get-txid-out-id

markhdavid:
This creates a function on a transaction to give you a "TXID"-like value

### DIFF
--- a/src/Cosi-BLS/block.lisp
+++ b/src/Cosi-BLS/block.lisp
@@ -174,11 +174,23 @@
 
 
 (defun get-txid-out-id (transaction)
-  "Get the identifier of TRANSACTION to hash at the leaves of the merkle
-   tree. In Bitcoin this is the TXID of each transaction, a hex string of length
-   64. Here, it is H(PubKey, PedComm)."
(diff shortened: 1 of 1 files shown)


src/Cosi-BLS/block.lisp
```diff
@@ -174,11 +174,23 @@
...
+  (vec-repr:bev-vec (hash:hash-val (hash-transaction transaction))))
+
+;; In Bitcoin this is known as the TXID of the transaction.
+
+
+
+(defun hash-transaction (transaction)
+  "Produce a hash for TRANSACTION, including the pair (PubKey, PedComm).
+   The resulting hash's octet vector is usable as a transaction ID."
+  (hash/256d transaction))
```
dbmcclain, line 190: I assume you have a definition of your own called HASH/256D ? The one I provide does not have a trailing "D", just HASH/256

dbmcclain (approved):

src/Cosi-BLS/block.lisp
markhdavid: yes I do (it's in the same file)

PR #103, opened 2018-04-19
[#15692472] added tar to end of production.bash

guitarvydas:


etc/production.bash
```diff
@@ -138,6 +138,14 @@ tar xf libs.tar
 cp ${etc}/emotiq.bash.${kind} ${production_dir}/emotiq.bash
 cd ${lib}
 rm *
-# < /test >
 
-# cp ${???}/*.so* ${production_dir}
+cd ${production_dir}
+case ${kind} in
+    linux) 
+	tar cfj ${kind}.emotiq.bz2 em* *.so*
```
easye: Please use an explicit manifest for archiving items rather than glob patterns for artifacts present on the filesystem.   

Otherwise, the distribution process will possibly pick up stray files.

easye: This is an unusual naming pattern for software archives.  

Please name the resulting archive
```
emotiq-${version}-${kind}.tar.bz2
```
to indicate that this is a tar archive.  

For `version` I would either use a timestamp of sufficient precision (minutes) to be unique, or read the contents of a file in git, or interogate the branch name.  You may want to ask Vlad/Eugene for guidance on how we wish to version our releases.

`kind` would be better named `arch`.

easye: It is unexpected for tar archives to untar their contents into the current directory. 

It is more usual for a given  archive to contain their contents contained in top-level directory within the archive.  

so an untar of `emotiq-20180420a-linux.tar.bz2` would produce
```
./emotiq-20180420a-linux/emotiq
./emotiq-20180420a-linux/emotiq.bash
./emotiq-20180420a-linux/libgmp.so
…
```

easye (changes requested):
Please see comments for suggested changes.  You can try convincing Vlad/Eugene that my suggestions are not necessary, but they are what I would expect of a basic software release mechanism.

etc/production.bash
guitarvydas: I can name each of the files explicitly.  Is there something better?

guitarvydas: OK to both.  I will change this stuff.

guitarvydas: This sounds like "best current practice".  I'll make the changes.

easye: The better way would be to follow DRY (Do not repeat yourself) whereby you record the contents which you move into the production directory, and use this to drive the process.

easye:
@guitarvydas Please ensure that the command line arguments given to `emotiq.bash` are passed to `emotiq` for parsing. 

I forgot to add this in the original review, but this is perhaps the most important change, because otherwise we cannot use command line arguments.

easye:
@guitarvydas 
Running the current version of `production.bash` under macOS fails:
```+ tar cfj emotiq--macos.bz2
tar: no files or directories specified
```
as you don't have a definition of libs in the "Darwin" clause.

easye:
@guitarvydas In the DRY (Do not Repeat Yorrself) department, I wouldn't duplicate the build instructions in `create-crypto-pairings.bash` in `production.bash`, but rather have `production.bash` invoke `create-crypto-pairings.bash` and *then* assemble the artifacts.  Otherwise maintainence is going to get complicated.

PR #104, opened 2018-04-19
Dbm 0418b

dbmcclain:
patch around unfinished code


guitarvydas (approved):
For "readability" by others, I would edit out the "#1=" stuff.  There appear to be only four #1#'s and these can be expanded to say (return-from check-block-transactions ...).  This can be done easily with a "replace string" command.

PR #106, opened 2018-04-20
add witness slots to block and various refinements and corrections

markhdavid:
Replace validator-keys-joining and validator-keys-leaving slots with three new slots: public-keys-of-witnesses, witness-bitmap, and block-signature. Moving towards preference for witness over validator as the term of choice. A couple of methods, ith-witness-signed-p and set-ith-witness-signed-p, abstract the bitmap implementation details. 

Fixed deficiency that TRANSACTIONS slot, although specified as a sequence, was assumed to be a list in compute-merkle-root-hash; now that is fixed. Actually, it would be preferrable for this to be a vector in most cases, but let's leave it open for now.

Hash-block needed to be updated to account for the witness slots. I believe hashing only the signature should be sufficient. Also decided in consultation with David that the serialization could be simplified by using lisp object encoder (loenc) on slots and no use or need for counting number of octets in the serialized input (c.f., Bitcoin's '80 byte header').

Removed comment about switching to Hash160 for TXID-like hash, since it was mistaken: that's not the hash used for TXID's in Bitcoin.

### DIFF
--- a/src/Cosi-BLS/block.lisp
+++ b/src/Cosi-BLS/block.lisp
@@ -11,7 +11,7 @@
     :accessor protocol-version
     :initform 1
     :documentation 
-      "Version of the protocol/software, an integer")
+      "Version of the protocol/software, an integer.")
 
    (epoch                               ; aka "height"
     :initarg :epoch :accessor epoch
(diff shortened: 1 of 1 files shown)


dbmcclain (approved):

dbmcclain:
We aren't quite there yet... some things will change, other things, such as setting bits in the bitmap are never done that way (See src/Cosi-BLS/cosi-handlers.lisp NODE-COSI-SIGNING where the inner routine FOLD-ANSWER produces a combined signature and simply LOGIOR's the bitmaps together.)

We'll need to hash everything in the block, except for the slot that contains this resulting hash value. Doing anything less, allows an attacker to replace pieces of the puzzle.

markhdavid:
@dbmcclain certainly there is no scenario in which the contents of the transactions slot would need to be hashed in addition to the merkle tree root, since the merkle tree root is by definition a hash of the contents of the transactions slot.

PR #116, opened 2018-04-28
157169441 renable lispwork ci tests

easye:
See <https://www.pivotaltracker.com/story/show/157169441>.


markhdavid (approved):
great, thank you

PR #120, opened 2018-04-30
Dbm 0423

dbmcclain:
cleans up confusing mess of package dependencies down in Crypto lib. Invented new package :cached-var, and removed unneeded :crypto-mod-math dependencies. Now load gives no undefined symbol refs, even in unused portions. (at least for Emotiq purposes)


easye (changes requested):
Please undo merge of `svs4` with this request.  I don't think Shannon's code is ready for `dev` quite yet, and it is not needed for the changes you have made to Cosi-BLS

PR #122, opened 2018-04-30
rewrite Actors =defun to elide nested backquotes

dbmcclain:


easye (approved):
Tested via running `cosi-simgen::tst-blk` under both `ccl` and `lwpro`.

PR #123, opened 2018-05-01
156854872 cli version of full node

guitarvydas:
(1) fixed tst-blk() to send startoff command to (node-self *top-node*)
(2) fixed linux pathname for ~/.emotiq directory (removed leading slash)


easye (changes requested):
I don't understand why this is broken up differently from <https://github.com/emotiq/emotiq/pull/124>?  What is the point of the two pull requests?  I don't think one can apply both #124 and this one to the tree.

PR #124, opened 2018-05-01
157210331 simulation node on one machine

guitarvydas:
README should contain instructions for simulation build.
Remember to change the IP addr in node-sim.lisp
I've tried to make it so that we need to specify the IP addr in only one place.
I've rebuilt the PR from scratch, leaving out unfinished items.


src/README
```diff
@@ -0,0 +1,20 @@
+
+to build and run:
```
easye, line 2: @guitarvydas As far as I can tell, the code referred to by `emotiq:node` doesn't do anything at this point.  Why should we be documenting it?

src/node-sim.lisp
```diff
@@ -0,0 +1,13 @@
+;; top-level glue code for a simulation node - enode == Emotiq Node
+ 
+(in-package :cosi-simgen)
+
+(defun node ()
```
easye, line 5: The user should not be expected to change Lisp code to run an entry function.  Better to parameterize the IP address as an argument to this function.  

I believe one could simply use 127.0.0.1  as the default value.

src/README
```diff
@@ -0,0 +1,20 @@
...
+You should see an "emotiq>" prompt, at which you can type 'help'.
+
+
+
+To get the simulator to work on a single machine:
+
+change the IP address in node-sim.lisp
+
+> (ql:quickload :emotiq/sim)
+> (cosi-simgen:enode)
```
easye, line 16: There is no symbol `cosi-simgen:enode`.  There is a `cosi-blkdef:enode`

easye (changes requested):

easye:
I would rather write this not to use the `tst-blk` code directly, as that it is not necessarily a stable interface, but rather use that code as an example to write our own logic orchestrating David's functions.  As far as I can see `tst-blk` only produces a single block, and then "runs off the rails".  Instead, we should continue to produce blocks in the background, as well as have the elections run for new blocks.

PR #125, opened 2018-05-02
Svs5

svspire:
Stable enough for David's use. Major bug fixed.


src/gossip/gossip-startup.lisp
```diff
@@ -0,0 +1,34 @@
+;;; gossip-startup.lisp
+;;; 21-Apr-2018 SVS
+;;; How to start up a node that can participate in gossip
+
+(in-package :gossip)
+
+(defparameter *default-config-path* "/usr/local/gossip/config" "Path to default config file")
```
easye, line 7: @svspire Probably better not to write configuration files outside the source tree.  Such a location may not be writeable by the current user, and on shared machines (like our AWS instances) it is unclear who should "control" such a location 

```
(defparameter *default-config-path* 
   (let ((config-dir (asdf:system-relative-pathname :gossip "conf/")))
      (ensure-directories-exist config-dir)
       (merge-pathnames "gossip.config" config-dir)))
```

"Path to default config file")

easye (approved):

PR #126, opened 2018-05-02
Dbm 0430a - add message authentication to gossip protocol

dbmcclain:


svspire (approved):
Everything still works. Tests pass and remote operation on AWS works.

PR #127, opened 2018-05-03
New directory. Proposal for startup configuration files.

svspire:
Please review and comment.


easye (commented):
This work ignores the wallet interface already present in the codebase at <https://github.com/emotiq/emotiq/blob/dev/src/wallet.lisp>.

easye (commented):
This configuration mechanism would need to refer to a filesystem abstraction independent of the source tree in order to work with binary-only distributions of Emotiq.

easye:
I worked the configuration code to a compilable state as <https://github.com/emotiq/emotiq/tree/config-mte>

easye:
@svspire: Is this pull request still to be actively considered for review?  If you no longer wish it considered, please close it.

PR #128, opened 2018-05-03
Dbm 0430a

dbmcclain:
add fees to transactions, make subkeys capable of signing transactions, add provisions for gossip signing in Cosi network

### DIFF
--- a/src/Cosi-BLS/cosi-handlers.lisp
+++ b/src/Cosi-BLS/cosi-handlers.lisp
@@ -1152,11 +1152,12 @@ bother factoring it with NODE-COSI-SIGNING."
                  (multiple-value-bind (utxo1 secr1) ;; sends
                      (make-txout 750 pkeym)
                    (multiple-value-bind (utxo2 secr2)
-                       (make-txout 250 pkey)
+                       (make-txout 240 pkey)
                      
                      (let ((trans (make-transaction `(,utxin) `(,info)
                                                     `(,utxo1 ,utxo2)
-                                                    `(,secr1 ,secr2))))
+                                                    `(,secr1 ,secr2)
+                                                    10)))
                        
                        ;; send TX to all nodes
                        (send-tx-to-all (setf *trans1* trans))
@@ -1174,11 +1175,12 @@ bother factoring it with NODE-COSI-SIGNING."
                            (multiple-value-bind (utxo1 secr1) ;; sends
                                (make-txout 250 pkeym)
                              (multiple-value-bind (utxo2 secr2)
-                                 (make-txout 500 pkey)
+                                 (make-txout 490 pkey)
                                
                                (let ((trans (make-transaction `(,utxin) `(,info)
                                                               `(,utxo1 ,utxo2)
-                                                              `(,secr1 ,secr2))))
+                                                              `(,secr1 ,secr2)
+                                                              10)))
                                  ;; send TX to all nodes
                                  (send-tx-to-all (setf *trans2* trans))
                                  ))))))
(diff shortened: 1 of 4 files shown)


dbmcclain:
Changes should be transparent to existing tests. Benign substrate changes.

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -775,6 +775,57 @@ check that each TXIN and TXOUT is mathematically sound."
...
       )))
 
+;; ------------------------------
+
+(defun gossip-neighborcast (node &rest msg)
+  "Gossip-neighborcast - send message to all nodes. Just a stub for now."
+  (declare (ignore node msg))
+  t)
+
+(=defun gossip-signing (node my-ip consensus-stage msg seq-id)
```
easye, line 785: @dbmcclain Doesn't the `my-ip` argument means that pseudo-nodes need to have to have a randomly generated ip address associated with them somehow?  I thought that this wasn't necessarily the case now.

src/Cosi-BLS/transaction.lisp
```diff
@@ -123,13 +123,16 @@ the simple commitment to the uncloaked value"
...
+            :initarg :fee
+            :initform 0)
    (gamadj  :reader  trans-gamadj  ;; A curve adjustment to get to zero balance.
             :initarg :gamadj) 
    ))
 
 ;; ---------------------------------------------------------------------
 
-(defun make-transaction (txins gam-txins txouts txout-secrets)
+(defun make-transaction (txins gam-txins txouts txout-secrets fee)
```
easye: @dbmcclain You need to change the method signature for `make-transaction` in `src/Cosi-BLS/t/cosi-tests.lisp`.  

Alternately, you could let current code continue to work via use of keyword arguments ala:

```
(defun make-transaction (txins gam-txis txouts txout-secrets &key (fee 0))
…
```

easye (changes requested):

src/Cosi-BLS/transaction.lisp
dbmcclain: thanks for catching that... I'm keeping fees as a required argument, instead of a default optional arg. Are there going to be any cases of fee-less transactions? I can change to your suggested form.

src/Cosi-BLS/cosi-handlers.lisp
dbmcclain: Nothing has changed in the simulation structure of the Cosi nodes yet. Still in preparation for merging with Gossip protocol. This new code simply provides a test of the ability to retain tree-like and/or gossip-like communications in Cosi signature gathering.

src/Cosi-BLS/transaction.lisp
easye: Naw, go ahead with what you are comfortable. 

I tend to abhor `&optional` stuff in favor of `&key` with default values for two reasons:

1)  Upgrading code with `&optional` arguments to add/subtract arguments is a real messy job.

2) After a function gets more than a couple arguments, I find that having  `&key` parameters gives a much better mnemonic for users of the API.   And allows one to specify sensible defaults.

src/Cosi-BLS/cosi-handlers.lisp
dbmcclain: The my-ip argument in this case refers to the leaders actual IP for use as a reply-to argument. So, actually, this is orthogonal to the issue of simulation. We need to tell each witness node who to send their partial signatures to. For now, I'm having the leader do all the coalescing.

easye (approved):
Let's go ahead and merge this.

src/Cosi-BLS/transaction.lisp
dbmcclain: yes, I can think of two cases for fee-less transactions. Those being when the fee is spent to participant witnesses or leader node, and when escrow amounts are forfeited. So we should probably change to an optional or, better, a keyword argument with default 0. I agree.

PR #129, opened 2018-05-03
initial commit to bring back block class (now as eblock class)

markhdavid:
fix class eblock name reference


dbmcclain:
Looks okay to me... no obvious problems. But the direction you are heading will probably need some adapter code written to help bridge the gap between the C world and Lisp. That's fine by me. These are application level architectural issues in which I don't feel expert enough to make comments beyond this.

dbmcclain (approved):

PR #130, opened 2018-05-04
Dbm 0430a

dbmcclain:
Remove all (DECLAIM (SAFETY ...)), gradual migration over to full Gossip in Cosi


svspire (approved):
Compiles and runs. CCL bug in loenc:decode has been fixed:
(setf rp (range-proofs:make-range-proof 123456789))
(loenc:encode rp)
(loenc:decode *)
Now runs without error.

easye:
N.b. the David McClain inspired tests work *IFF* one changes the `cosi-simgen::*local-nodes*` special to something that can resolve the hostname.  

For my machine (named `illin.local` in mDNS), I need the following form to get the tests to run.

```
(setf cosi-simgen::*local-nodes* '(("illin.local" . "127.0.0.1")))
```

PR #132, opened 2018-05-07
fixes to block create

markhdavid:


markhdavid:
closing, since as Mark E correctly pointed out, this is more of a work in process.

PR #134, opened 2018-05-09
[#157420655] Replaced wget with curl

echupriyanov:
Following Paul's comment, replaced `wget` with `curl`


easye (approved):
`curl` is in `/usr/bin/` under my macOS install so this will certainly be more widely available.

PR #138, opened 2018-05-09
minor change from (machine-instance) to localhost

guitarvydas:
This fixes the "localhost" vs. "(machine-instance)" that was caused when PR133 was merged into dev.

One should first confirm that the current version of 'dev' gets an assertion failure when running (cosi-simgen::cosi-generate :nodes 8) [the 3rd line of simulation.md].

The problem (at least on my machine) is that the hostname for simulation was changed to "localhost" in many place, but a call to (MACHINE-INSTANCE) remained (in cosi-config.lisp).

### DIFF
--- a/src/Cosi-BLS/cosi-config.lisp
+++ b/src/Cosi-BLS/cosi-config.lisp
@@ -6,7 +6,7 @@
 (defun cosi-init (leader-node-ip)
   "Steps necessary to initialize specials in the cosi-bls system to
    create a machine local simulation network."
-  (let ((machine-name (machine-instance)))
+  (let ((machine-name "localhost"))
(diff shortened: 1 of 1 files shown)


easye:
@guitarvydas This configuration is no longer necessary with the simulation, as it does a purely local version by default.  Unfortunately, this was documented correctly in the simulation documentation. 

I have corrected things on the <https://github.com/emotiq/emotiq/tree/mte-simulation> branch, which is what I would propose using rather than this pull request.

easye (changes requested):
I have worked through the configuration problems <https://github.com/emotiq/emotiq/tree/mte-simulation> which doesn't need any changes to `cosi-bls`, and therefore is a better example of the simulator performing integration across David's codebase.

PR #140, opened 2018-05-10
Mte simulation 20180510a

easye:
Aligns documentation with running the simulator.

Subsumes <https://github.com/emotiq/emotiq/pull/139> and <https://github.com/emotiq/emotiq/pull/138>.


src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -1078,7 +1083,11 @@ bother factoring it with NODE-COSI-SIGNING."
 
 ;; ------------------------------------------------------------------------------------------------
 
-(defun leader-exec (node)
+(defun leader-exec (node &key (cosi-sign-prepare-timeout 10))
```
guitarvydas, line 1086: Should probably be cosi-timeout (see below, I hope).  There are two hard-coded 10's in the original code.

Ah, I can't comment on code that you haven't changed... There is another hard-coded 10 in the code (ac:self-call :cosi-sign-commit self new-block 10) .  The 10 should be changed to cosi-timeout (I think - unless you did this on purpose).  Leader-exec sends itself "wait-prep-signing" which then gives itself a continuation and self-calls :cosi-sign-commit (with the 2nd hard-coded 10)

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -36,6 +36,10 @@ THE SOFTWARE.
 
 (defvar *current-node*  nil)  ;; for sim = current node running
 
+(defvar *cosi-sign-prepare-timeout* 10
```
guitarvydas, line 39: *cosi-timeout* (see below)

src/node-sim.lisp
```diff
@@ -2,15 +2,37 @@
  
 (in-package :emotiq/sim)
 
-(defun initialize (&key (leader-node "127.0.0.1"))
+(defun initialize (&key
+                     (executive-threads nil)
+                     (nodes 8)
+                     (new-configuration-p nil)
+                     (cosi-sign-prepare-timeout 10)
```
guitarvydas, line 9: cosi-timeout

src/Cosi-BLS/cosi-handlers.lisp
easye: @guitarvydas I changed the code to mimic what you had done in #139.   You only changed the first timeout.

PR #143, opened 2018-05-10
Bring back block class now as eblock class

markhdavid:
Tested using simulation.md instructions.  Things seemed to run fine.

I noticed very first block created is not being added to list returned by (emotiq/sim:blocks). Not sure why that is.  But saw at least three blocks created.  Their epoch counts were 1, 2, 3, respectively.  They looked OK.

### DIFF
--- a/src/Cosi-BLS/block.lisp
+++ b/src/Cosi-BLS/block.lisp
@@ -6,43 +6,42 @@
 
 
                                                                                 ;
-(defclass block ()
+(defclass eblock ()
   ((protocol-version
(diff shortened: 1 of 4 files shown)


dbmcclain (approved):
Checkout of this branch and clear .cache. 
(asdf :cosi-bls)
(in-package :cosi-simgen)
(init-sim)
(tst)
(tst-blk)

Both tests run to completion without error.

markhdavid:
I closed this due to various issues that came up in the review.

PR #144, opened 2018-05-10
Pt 10may18

guitarvydas:
cosi timers re-instated

### DIFF
--- a/src/Cosi-BLS/cosi-handlers.lisp
+++ b/src/Cosi-BLS/cosi-handlers.lisp
@@ -92,7 +92,7 @@ THE SOFTWARE.
       ;; for sim and debug
       
       (:make-block ()
-       (leader-exec node *cosi-timeout*))
+       (leader-exec node *cosi-prepare-timeout* *cosi-commit-timeout*))
 
       (:genesis-utxo (utxo)
        (record-new-utxo (bev (txout-hashlock utxo))))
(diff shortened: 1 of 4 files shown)


src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -1078,7 +1084,7 @@ bother factoring it with NODE-COSI-SIGNING."
 
 ;; ------------------------------------------------------------------------------------------------
 
-(defun leader-exec (node)
+(defun leader-exec (node prepare-timeout commit-timeout)
```
easye, line 1087: @guitarvydas I would rather see additional arguments beyond the first, primary one be keyword arguments with the defaults being the package special, i.e. 
```(defun leader-exec (node &key (prepare-timeout *cosi-prepare-timeout*) (commit-timeout *commit-timeout*))```
This provides more clues in the source how these variables are being used, allows `leader-exec` to be called without having to remember anything about the order of the parameters beyond the first one.

File: src/node-sim.lisp [hunk 2]
```diff
@@ -3,6 +3,8 @@
 (in-package :emotiq/sim)
 
 (defun initialize (&key
```
easye, line 5: @guitarvydas With so many variables being used in `intitialize` it is probably a good idea to have more of an API for setting and inspecting the current state of how the simulator is set up to run.  But let's see how the coming configuration for real network transport shakes out (i.e. absorbing Shannon's TCP networking configuration mechanism) before extending this.

easye (approved):
Works for me.

PR #145, opened 2018-05-11
Bring back block class now as eblock class

markhdavid:
Let's try this now. This improves on the earlier abortive PR. The block is created accepting witnesses, leader-pkey, and leader-pkey.

### DIFF
--- a/src/Cosi-BLS/block.lisp
+++ b/src/Cosi-BLS/block.lisp
@@ -6,43 +6,48 @@
 
 
                                                                                 ;
-(defclass block ()
+(defclass eblock ()
   ((protocol-version
-    :accessor protocol-version
     :initform 1
     :documentation 
       "Version of the protocol/software, an integer.")
 
    (epoch                               ; aka "height"
-    :initarg :epoch :accessor epoch
-    :initform nil 
+    :initarg :epoch
+    :initform 0
     :documentation
-      "The integer of the epoch that this block is a part of."
-      ;; height: "Position of block in the chain, an integer."
-      )
+    "Increasing integer count of epoch this block is part of. We keep
+     a single chain combining identity and ledger, unlike Byzcoin,
+     which keeps an outside identity blockchain with its epoch count
+     and an inside ledger blockchain with its height count. Thus,
+     epoch subsumes height.")
 
-   (block-hash
-    :accessor block-hash
-    :documentation "Hash of this block.")
    (prev-block-hash
-    :accessor prev-block-hash
+    :initform nil
     :documentation "Hash of previous block (nil for genesis block).")
 
-   (block-timestamp
-    :accessor block-timestamp
+   (timestamp
     :documentation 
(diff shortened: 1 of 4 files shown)


markhdavid:
Ran the dev version (with the bc-block struct), and pushed it along to having 5 blocks. Did the equivalent in this branch.  Comparing these, they looked about the same.  The bitmaps were not identical.  Is that a concern?  Or is that semi-random.  Here's a collection code and the result from the latest runs:
```
;; eblock class:
(loop for b in (reverse (emotiq/sim:blocks))
               collect (slot-value b 'cosi/proofs:signature-bitmap))
(2 15 15 15 2)

;; bc-block structure:
(loop for b in (reverse (emotiq/sim:blocks))
               collect (slot-value b 'cosi-simgen::signature-bitmap))
(4 15 15 15 15)
```

markhdavid:
OK, regarding bitmap discussion from above note, I got a randomly different result on dev (using old bc-block structure):
(2 3 2 3 3)
So, this answers the question: no, these need not be identical.

easye (commented):
Need to update `*names-of-block-slots-to-serialize*` with all fields.

src/Cosi-BLS/block.lisp
```diff
@@ -6,43 +6,48 @@
 
 
                                                                                 ;
-(defclass block ()
+(defclass eblock ()
```
easye, line 9: Note need to sync with `*names-of-block-slots-to-serialize*` in a comment.  Possibly place nearby in source.

easye (commented):
Rename `blockchain.lisp` to something to indicate it is not the current block definition.

easye (changes requested):
Please:

1) update the serialization fields; note the need to synchronize them with the `defclass`

2) Mark `src/blockchain.lisp` as not the main block definition (move to a subdirectory called `brooks`?)

3) Add accessors to the `eblock` defclass

src/Cosi-BLS/block.lisp
easye: @markhdavid A more forward looking strategy would be to use `closer-mop` to just collect the slot names:

```
(loop :for slot :in (clos:compute-slots (find-class 'cosi/proofs:eblock))
   :collecting (slot-value slot 'clos::name))
```

markhdavid:
I've added the missing slots and moved the definitions closer to one another and documented the need to keep them in sync. CLOS access is a much more complicated approach, not sure it's worth it. I'll review it at a later time. 

Blockchain.lisp is in a completely different system from this. It should be reviewed, but there are several not-really-the-real-blockchain parts of the code base like that. It should not be part of this commit. 

I want to defer on the question of what readers or accessors to define for this class, as well as on a possible coding convention to define an accessor for every single slot of every single class.

markhdavid:
Actually want to close this, and open a new one on a new branch. Will advise soon.

PR #150, opened 2018-05-11
Pt sim election

guitarvydas:
It appears that faked elections work.  I believe that there are enough comments in src/election.lisp and node-sim.lisp.

### DIFF
--- a/src/Cosi-BLS/cosi-handlers.lisp
+++ b/src/Cosi-BLS/cosi-handlers.lisp
@@ -69,10 +69,6 @@ THE SOFTWARE.
       (:election (new-leader-pkey)
        (node-elect-new-leader new-leader-pkey))
 
-      (:hold-an-election (n)
-       (let ((tree (sim-trial-election:hold-trial-election n)))
-         (ac:pr (format nil "election results(~A) ~A" n tree))))
-
       ;; -------------------------------
       ;; internal comms between Cosi nodes
       
--- a/src/Cosi-BLS/cosi-handlers.lisp
+++ b/src/Cosi-BLS/cosi-handlers.lisp
@@ -68,7 +68,11 @@ THE SOFTWARE.
 
       (:election (new-leader-pkey)
        (node-elect-new-leader new-leader-pkey))
-      
+
+      (:hold-an-election (n)
+       (let ((tree (sim-trial-election:hold-trial-election n)))
+         (ac:pr (format nil "election results(~A) ~A" n tree))))
+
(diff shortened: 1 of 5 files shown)


src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -1316,10 +1316,12 @@ bother factoring it with NODE-COSI-SIGNING."
 
 ;; -------------------------------------------------------------
 
+#|
```
easye, line 1319: @guitarvydas Just go ahead and remove such references to a specific network configuration.

src/sim-election.lisp
```diff
@@ -0,0 +1,138 @@
...
+(defvar *beacon-interval*  7)
+(defvar *self-destruct*    nil)
+
+(defun kill-beacon ()
+  (setf *self-destruct* t))
+
+;; mp is the multi-processing library for LispWorks
+(defun make-trial-election-beacon (handler-actor)
+  (unless *beacon-timer*
+    (setf *beacon-timer* (mp:make-timer (lambda ()
```
easye, line 43: @guitarvydas We have to make the code work under `ccl` (and hopefully soon `sbcl`).  I think you should use the `mp-compat` system here to make a timer, or go find a suitable system in Quicklisp

easye (changes requested):

src/node-sim.lisp
```diff
@@ -31,9 +31,21 @@ witnesses."
               (not (and (probe-file cosi-simgen::*default-data-file*)
                         (probe-file cosi-simgen::*default-key-file*))))
     (cosi-simgen::generate-tree :nodes nodes))
-  (setf cosi-simgen::*cosi-prepare-timeout* cosi-prepare-timeout)
```
easye, line 34: @guitarvydas You've deleted the configuration options to set the Cosi timeouts.  Please reinstate, or explain why you deleted this option.

easye (changes requested):
1.  Elections are happening too fast for anything meaningful.  Please allow the configuration of the time between elections.

2.  Holding an election without sealing a block doesn't make sense.  When the election timer fires, the block should be sealed, then the election held.  As such, you should probably deprecate the `force 

3.  Please output the identity of the new leader after the election has occurred.

dbmcclain:
Elections should be run by all witness nodes. There is no need to broadcast the outcome of the leader election. If, perchance, two different nodes think they both won the election, only one - at most - can achieve BFT.

PR #151, opened 2018-05-11
ignore Cosi-BLS/config

guitarvydas:
added to Cosi-BLS/config to the ignore list


easye (changes requested):
This is already part of `dev` <https://github.com/emotiq/emotiq/blob/dev/.gitignore#L17>

PR #154, opened 2018-05-14
Pt uncloaked4

easye:


guitarvydas:
This appears to produce blocks with 0 transactions...

PR #155, opened 2018-05-15
Socket actors

svspire:
TCP connections are working on CCL both between local processes on Darwin and between Darwin processes and AWS.


lukego:
I am new to this codebase so my review is to report my understanding of what this branch does and what happens when I run it.

### Understanding

My high-level understanding is that we have two layers here. The bottom layer is Actors which provides message passing *within* the Lisp image. The top layer is Gossip which adds distributed operation over a network by running certain actors that own sockets and know how to route messages.

(This is a different formulation than Erlang where the bottom layer handles distributed message routing and every actor's *PID* is addressable from any node, but it has the same effect.)

The blockchain functionality will then be a third (etc) layer built on top of Gossip and therefore able to route messages between nodes.

Is that correct?

### Testing

I have tested this code by running two CCL processes each loading `socket-actors-test.lua`. In the first process I run `(gossip::loopback-server)` and in the second process I run `(gossip::test-sa-client1)`. I also use `TRACE` to see certain mailbox and actor calls. Overall I am able to see the message delivered from the client to the server, implying that the socket is setup and working, but I see an error on the client when it tries to receive a reply. 

The exact error is:

```
There is no applicable method for the generic function:
  #<STANDARD-GENERIC-FUNCTION MP-COMPATIBILITY::DEQUEUE #x302001A2E6DF>
when called with arguments:
  (GOSSIP::OFSAG 10.0)
   [Condition of type NO-APPLICABLE-METHOD-EXISTS]
```

and this seems to be caused by `ENSURE-CONNECTION` calling `MAKE-SOCKET-ACTOR` with the symbol `OFSAG` where an `MPCOMPAT` mailbox is expected. I have the impression that this symbol is fbound to a callback function that should handle incoming messages but I am not sure about that.

Is this a bug? I must admit that I have merged `dev` into this branch for testing (for access to some of the new QuickLisp setup code) and this may have introduced a problem (but presumably that would need to be resolved for merge anyway.)

lukego (changes requested):
I was able to run this successfully except for the very last step, where the client tries to retrieve a reply from its mailbox. I would like to understand: Is this a bug?

I also have a workflow question: Since this code is currently only running on CCL and not LispWorks (according to Travis) does that mean that it needs to be ported before merge onto `dev`?

dbmcclain:
I think it is a mistake to view Actors vs Gossip as "layers". Erlang has its Processes with an underlying message passing system. Our Actors offer similar behavior with SEND. Gossip is a mechanism for transmission across networks. It can be constructed, in part, from Actors, other parts from dedicated server Threads.

Erlang extends its own SEND down in the kernel, to work across networks, and we can do that too. But that isn't really a function of Actors in our system, or Processes in Erlang. Rather, it is a layer of message delivery atop which Actors and Processes can be extended to work. In other words it is an additional DEFMETHOD overloading on SEND.

About the only semantic difference between our Actors and Erlang is with respect to Erlang LINKS. We don't have those, and in 10 years experience with Butterfly(*) I have rarely seen them as necessary, simply because Lisp has a powerful condition system that goes far beyond anything that Erlang could offer.

So I view Gossip as a networking protocol. I view Actors as task management. The two are orthogonal, and yet can take advantage of what each has to offer. Nothing hierarchical about this relationship.

(*) Butterfly started out as Thread-based (heavy) Erlang in Lisp, complete with LINKS. It has since been adapted to Actors as well. But apart from capacity for running "processes", original Butterfly was equivalent to Erlang, including transparent cross-platform message delivery. I rarely needed the added complexity of LINK out-of-band communications, given that I had a Lisp Condition System and UNWIND-PROTECT, HANDLER-BIND, etc. that avoided the need to kill/restart a network of linked processes. No reason to kill a process on trivial errors.

lukego:
Thanks for explaining @dbmcclain!

lukego:
@svspire I'm running the tests in `network-test.lisp` too after setting `*server-address*` to `"localhost"`. Superficially this all seems to work nicely in that the commands execute without errors and they produce pretty pictures like the ones attached. I'm not sure how to interpret this yet though.

Do we have any ways to automate multi-process testing? If not then would it make sense to set this up, or would that be premature for some reason?

screen shot 2018-05-15 at 15 58 34
screen shot 2018-05-15 at 15 31 36

src/config/config-file-utilities.lisp
```diff
@@ -0,0 +1,88 @@
+; notional code to read configuration files and generate
```
easye, line 1: @svspire Needs to be hooked up into an ASDF definition.  I provided code for you to use in <https://github.com/emotiq/emotiq/tree/config-mte/src/config>

src/config/gossip.conf
```diff
@@ -0,0 +1,12 @@
+; example gossip configuration file
```
easye, line 1: @svspire Is the intention to have a single global configuration committed to our source repository?  I think you would want these files to by examples, rather than universal configuration.

svspire:
`(gossip::loopback-server)` is old code and no longer works (but I probably should fix it).
`network-test.lisp` is the real test and yes, that needs to be automated somehow.

easye (changes requested):
I would like to see the configuration system actually working to configure nodes.

svspire:
The green nodes are "real" nodes on the local machine; the yellow nodes are proxy nodes that stand in for a real node on a different machine.

src/config/gossip.conf
svspire: It's all examples right now. Remove the whole config subdirectory if you wish. I hadn't intended to include it anyway.

PR #157, opened 2018-05-15
Pt 0430a

guitarvydas:
dbm-0430a merged with dev (no-brainer), simulation modified for transaction signatures


guitarvydas:
This PR introduces transaction signatures and re-vamped src/Cosi-BLS/transaction.lisp which should help with work on uncloaked transactions.

src/node-sim.lisp
```diff
@@ -2,6 +2,10 @@
  
 (in-package :emotiq/sim)
 
+;; remove for production
+(defun eassert (bool-condition)
```
easye, line 6: @guitarvydas I really don't like this sort of interface.  

If you are going to cause the code to signal an error, you need to provide restarts for the user to choose (which is the whole point of `cl:assert` interface and associated protocol).

If you are simply trying to provide a test that the code does what one does, then use a test framework to write a unit test.

Writing a note to "Change for production" is the kind of thing that almost always gets overlooked when production actually happens.  

So, either factor these assertions out into a test, or provide a meaningful restart for the user.  

Since this is currently in the simulator which isn't necessarily going to be in production, I'm not going to neg. this pull request, but please figure out if this assertion should be correctable or be a unit test and change when you get a chance.

src/Cosi-BLS/transaction.lisp
```diff
@@ -67,26 +86,40 @@ It should be stored in the wallet"))
...
   "We are often given a long Bulletproof, and we need the to hash on
 the simple commitment to the uncloaked value"
   (hash:hash/256 (pedersen-commitment prf) pkey))
 
-(defmethod make-hashlock ((amt integer) (pkey pbc:public-key))
-  (hash:hash/256 amt pkey))
+(defmethod make-hashlock (arg (pkey pbc:public-key))
+  (hash:hash/256 arg pkey))
+
+(defun make-admin-commit (amt &optional (gamma (random-between 1 *ed-r*)))
```
easye, line 105: @dbmcclain What is an "admin commit"?

src/Cosi-BLS/transaction.lisp
```diff
@@ -67,26 +86,40 @@ It should be stored in the wallet"))
...
+  (values
+   (range-proofs:simple-commit (range-proofs:hpt) gamma amt)
+   gamma))
+
+(defun check-amt (amt)
+  (and (not (minusp amt))
+       (< amt #.(ash 1 64))))
```
easye, line 112: @dbmcclain:  Where does the maximum amount of `(ash 1 64)` come from?  The simple need to have an upper bound?  Or is this related to some sort of serialization constraint?  Or is this the maximum that the range proofs can contain?

src/package.lisp
```diff
@@ -75,7 +75,9 @@
 
    
    #:*user-1* #:*user-2* #:*user-3*
-   #:*tx-1* #:*tx-2* #:*tx-3*))
+   #:*tx-1* #:*tx-2* #:*tx-3*
+
+   #:eassert))
```
easye, line 80: @guitarvydas Why export this symbol?  Do you intend to make an API for others to use?

easye (approved):
It would be better to explain what is going on in commits in a coherent manner, but this request is no worse in explanation/coherency than others.

easye:
@guitarvydas Do you actively want this pull request anymore?  It has been approved, so it could be merged if you wanted to.

PR #159, opened 2018-05-17
Merge working gossip code into dev 

easye:
* completed migration to TCP
* configuration code works 
* tested with both LispWorks and CCL


src/Actors/actors-startup.lisp
```diff
@@ -214,16 +216,17 @@ THE SOFTWARE.
   (install-actor-directory)
   (install-actor-printer))
 
-#+(or :ALLEGRO :OPENMCL)
-(install-actor-system)
+#+(or :LISPWORKS :ALLEGRO :OPENMCL)
```
easye, line 219: Why is there a conditional?  Shouldn't this be for all platforms?

src/gossip/socket-actors-test.lisp
```diff
@@ -0,0 +1,55 @@
+;;; socket-actor-test.lisp
```
easye, line 1: @svspire Maybe a comment to mark status as not working.

easye:
Copy config files to 
```
(asdf:system-relative-pathname :emotiq "../var/etc/gossip-config/")
```

src/Actors/actors-startup.lisp
svspire: I'm going to leave the conditional here for now until we can ensure that all Lispworks images have multiprocessing enabled. The possibility that that isn't happening is why the old conditional for lispworks is still present but commented out.

svspire (approved):
Travis deploy tests are failing because they seem to be using a non-multiprocessing version of Lispworks. I'm going to go ahead and approve and merge with dev, because the fix for Travis is outside this code.

PR #160, opened 2018-05-19
Dbm 0519a

dbmcclain:
merged dev branch into this branch as of 21:38 UTC on May 19, 2018. 

Fix block validation to perform proper block hash checking between transactions and stored hash in block header, transaction validation, during Cosi PREPARE phase. 

Fix block header hash calculation to exclude block signature slots (sig, pkey, bitmap). 

Perform full multisignature validation, check that composite signing public key is crypto sum of witness keys as indicated in the signature bitmap, during Cosi COMMIT phase.


dbmcclain:
Fix node message dispatch to use CLOS method with EQL dispatch on first argument (typically, a symbol message name). Fix Emotiq/sim to conform to new dispatch mechanism. No need to override existing handlers, unless you want to. Cosi and Emotiq sim tests now function together without any clashes. 

Emotiq/sim tests on cloaked and uncloaked transactions runs okay now. Fixed election sim to likewise conform to new message dispatch.

Extended exports from cosi-simgen to alleviate references to private symbols in Emotiq sim.

src/Cosi-BLS/cosi-netw-xlat.lisp
```diff
@@ -51,6 +83,30 @@
 
 (defmethod unregister-return-addr ((ret return-addr))
   (unregister-aid (return-addr-aid ret)))
+|#
+;; -----------------------------------------------------------------
+
+(defvar *machine-ip-addr* (comm:get-host-entry (machine-instance) :fields '(:address)))
```
easye, line 89: @dbmcclain Referencing the LispWorks only `comm` package causes compilation to fail on other implementations.

src/Actors/actors-startup.lisp
```diff
@@ -213,20 +213,20 @@ THE SOFTWARE.
...
 (defun install-actor-system (&rest ignored)
   (declare (ignore ignored))
-  (install-actor-directory)
-  (install-actor-printer))
+  (unless (directory-manager-p)
+    (install-actor-directory)
+    (install-actor-printer)))
 
-#+(or :LISPWORKS :ALLEGRO :OPENMCL)
+#+(AND (NOT :COM.RAL)
```
easye, line 220: @dbmcclain Why do we want to conditionalize on running on these Lisps?  Is this related to some need of the Actor code?  Shouldn't you just attempt to `install-actor-system` for all Lisps from this point, and let the `actor` system deal with whether it can run on a given implementation?

src/node-sim.lisp
```diff
@@ -151,187 +147,66 @@ This will spawn an actor which will asynchronously do the following:
...
-|#
 
+  (cosi-simgen:reset-nodes) 
+
+  (emotiq/elections:make-election-beacon)
+                                         
+  ;(ac:spawn  ;; cosi-handlers assumes that there is one block in the blockchain, if you spawn here,
+  ; you might get errors in cosi-handlers if leader-exec runs before this code...
+  ; (lambda ()
+  (let ((fee 10))
```
easye, line 165: @dbmcclain If you are going to change what the function does and it has a docstring, you need to change the docstring to reflect the new behavior.

src/simulation.md
```diff
@@ -4,7 +4,7 @@
     
     (ql:quickload :emotiq/sim)
     (emotiq/sim:initialize)  ;; takes several keywords - see node-sim.lisp
-    (emotiq/sim:run)
+    (emotiq/sim::run)
```
easye, line 7: @dbmcclain The symbol is exported, so no need to document calling the private symbol.

src/Cosi-BLS/range-proofs.lisp
```diff
@@ -219,7 +219,404 @@ THE SOFTWARE.
 (define-symbol-macro *curve*   (bp-basis-curve  *bp-basis*))
 (define-symbol-macro *nbits*   (bp-basis-nbits  *bp-basis*))
 
-(defvar *bp-basis* 
+#| ;; for 192 bit security...
+(defvar *bp-basis*
+  (make-bp-basis :CURVE :CURVE-E382
```
easye: @dbmcclain Where are these constants coming from?

easye (changes requested):
Please resolve issues with @guitarvydas.

src/Actors/actors-startup.lisp
dbmcclain: You may not want to conditionalize on Lisp variant. But for general purpose use, I need my copy of Lipspworks to use the ACTIONS system to fire up Actors. These Actors are used in my lab system for running lab equipment and need to get fired up at init time, but only after multiprocessing has started. Hence use of (NOT :COM.RAL)

src/simulation.md
dbmcclain: I didn't touch this file. I don't know what its purpose is. I merely touched up Paul's sources to remove references to cosi-simgen internal symbols.

src/Cosi-BLS/range-proofs.lisp
dbmcclain: The constants are generated at genesis time (by me, in the source code) to establish basis vectors in the Elliptic curves for use in Bulletproofs. Lacking the pre-made basis vectors, we have the system auto-initialize, but that bites on every Bulletproof. If they pre-exist, then we save the time needed to generate the basis vectors in every call to MAKE-RANGE-PROOF.

So, to answer your question, I find the values by first setting *bp-basis* to nil, then run (MAKE-RANGE-PROOF 15) and then PPRINT the value of *bp-basis* in the REPL and copy back into the source file. (the value 15 is just a PIDOOMA - can be any integer value in range)

src/Cosi-BLS/cosi-netw-xlat.lisp
dbmcclain: Yes, I understand. Perhaps our network experts know a better way? But at any rate this particular code disappears once I am fully hooked into Shannon's Gossip system.

src/node-sim.lisp
dbmcclain: I didn't change the semantics of anything, as far as I know, just syntactic substitutions.

PR #161, opened 2018-05-23
Dbm 0521a

dbmcclain:
add subkey-derivation.lisp to repo along with all other changes up to this point


easye (approved):
`emotiq/sim` seems to be broken under `ccl` due to rentrancy problems, but approving anyways to get forward motion on `dev`.

PR #162, opened 2018-05-23
Add initial networking benchmark suite socket-actors-bench.lisp

lukego:
This file defines a simple transport-level benchmark called
'sendrecv-tcp-1' that opens a set of client-server socket pairs and
sends messages from the client to the server.

The basic benchmark function GOSSIP::BENCHMARK-SENDRECV-TCP-1 takes
keyword parameters to control how the test is executed. This is
especially meant to test the performance of different socket polling
strategies, message serialization methods, etc, with different levels
of load.

The higher-level function GOSSIP::SUITE-SENDRECV-TCP-1 can execute a
complete test campaign for the permutations of given variable values.
The results are formatted as CSV for separate analysis.

Notes:

- This benchmark is operating at the socket layer and does not
  exercise higher-level functionality from socket-actors and gossip.
- The benchmark runs all clients and servers on a single Lisp thread.
- This initial implementation has not yet been sanity-checked with
  tools like system call profilers and so may contain egregious bugs.

### Testing

This is easy to test:
- Load the file `gossip/socket-actors-bench.lisp`.
- Run `(gossip::suite-sendrecv-tcp-1)`.

You may need to check your `ulimit -n` for maximum open file descriptors if you are testing with many sockets at once.

### Example

Here is an example set of results in CSV format: [gist](https://gist.github.com/lukego/2fdba52a042aaaf5f83ac7b02686b0f7). These are from a campaign running on CCL/Linux that tested many different levels of socket concurrency and many different levels of "chattiness" to see how this affects throughput and how much difference the polling method makes (comparing only two basic polling methods, NOWAIT that simply receives on every socket in a loop, and USOCKET that calls `select()` under the hood.)

The CSV results can then be analyzed in weird and wonderful ways. Here is one (click to zoom):

rplot15

Here is one interpretation:

- The panels each show a different level of "chattiness" from 0.1 where 10% of sockets have data on each iteration to 1.0 where all sockets have data.
- The Y axis shows throughput in total messages/second transferred across client/server pairs.
- The X axis shows the number of client/server socket pairs that are open concurrently.
- The color of each point shows the polling method that was used.
- The slope from left to right shows how performance changes (degrades) as more sockets are opened.
- The points exhibit some jitter. Each point in this dataset represents one second of runtime and this can be significantly impacted by background events such as garbage collections.

This graph seems to suggest that with usocket on CCL/Linux performance falls fairly gradually as the number of sockets increase, with relatively little difference between 1 socket and 100 sockets, and that there is not much practical difference between these polling methods at the measures levels of chattiness.

Quite a few data points are down ~10-20% seemingly at random. My hypothesis is that this is due to gabarge collection. In the future this could either be controlled for explicitly, or averaged out with longer tests, or simply ignored.

### DIFF
--- a/src/gossip/socket-actors-bench.lisp
+++ b/src/gossip/socket-actors-bench.lisp
@@ -6,13 +6,13 @@
 
 (in-package :gossip)
 
-(defun benchmark-sendrecv-tcp-1 (&key (n 1) (seconds 1) (msgsize 100)
+(defun benchmark-sendrecv-tcp-1 (&key (connections 1) (seconds 1) (msgsize 100)
                                    (chatty 1.0)
                                    (encode :raw)
                                    (wait :nowait)
                                    (address "localhost"))
   "Run a benchmark of N client-server pairs sending unidirectional messages.
-   N is the number of client-server socket pairs to establish.
+   CONNECTIONS is the number of client-server socket pairs to establish.
    SECONDS is the duration of the benchmark.
    CHATTY is the chance of a socket being active on each iteration (0.0 - 1.0).
    ENCODE is :RAW or :SERIALIZE for how the message is encoded.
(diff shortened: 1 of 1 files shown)


lukego:
Note: The exact code that produced the CSV above was:

```lisp
(suite-sendrecv-tcp-1 :n (loop for n from 0 to 1000 by 25 collect (max n 1))
                      :msgsize '(1000)
                      :chatty '(0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.0)
                      :encode '(:raw)
                      :output-filename "results.csv")
```

lukego:
I pushed an update to this benchmark with commit 86a0517. The `USOCKET` wait mode was not really being tested so we didn't see its true cost and now I added a `THREAD` wait mode that spins off a separate thread to receive each message (but still always sending with just one thread.)

Here is another example of using data from the benchmark, this time to see the trend in performance from 1..500 sockets for different polling methods at four different levels of "chattiness." Click to zoom:

rplot18

(Not pictured: CCL process reliably crashing if the threads limit is raised to 1000.)

markhdavid (changes requested):
Hi Luke, I wanted to add my review. Overall, this is really great looking code and documentation.  However, there's one thing I care about that I would like you fix first.  That is that the advertised functions have not been exported.  Do you think you could change BENCHMARK-SENDRECV-... and any others. Generally, ```<package name>::<symbol>``` forms are a red flag. Nobody should be calling functions that way - if they need to call it, it should be exported. Thank you!

lukego:
Thanks for the review @markhdavid. I reckon that I will close this PR now and fold the benchmark code into the new GOSSIP/TRANSPORT module/package that I am working on now. I reckon that these symbols will look more appropriate being exported from there rather than the main Gossip package.

PR #164, opened 2018-05-23
Dbm 0523a

dbmcclain:
Fix only the problematic :around method in cosi-netw-xlat.lisp that caused CCL runtime errors.


easye (approved):
Verified as working under CCL.  Thanks!

PR #167, opened 2018-05-25
Delete war-and-peace.htm

dbmcclain:


easye (approved):
A loss for literature, but a gain for code statistics.

PR #169, opened 2018-05-27
bitcoin-style base58-check p2pkh addresses

markhdavid:
Adding a facility to hash a public key. We do this in one of the most standard ways done by several cryptocurrencies, namely, to use Bitcoin's Base58Check encoding but have our own version ID (also known as a network byte). Details are in the doc and code, and are also pretty well documented in Bitcoin and other cryptocurrency forums and specs.  I implemented a basic barebones testing facility. (Because this is so standard, it's convenient that you can also use public testing sites to test this against, and there's doc in the sources to point you to at least one.) The main export is function public-key-to-address in package cosi/proofs.

### DIFF
--- a/src/Cosi-BLS/address.lisp
+++ b/src/Cosi-BLS/address.lisp
@@ -1,0 +1,210 @@
+;;;; address.lisp
+
+(in-package :cosi/proofs)
+
+
+
+;;;; Addresses: Public Key Hashes in Emotiq
+
+;;; Hashed public keys are presented as `addresses'.  An `address' for a public
+;;; key is in general produced as follows:
+;;;
+;;; First, it is hashed irreversibly, and then encoded in a user friendly
+;;; manner, that is, shorter; with characters easily typed, read, and
+;;; distinguishable by humans in nearly all computing environments; with a
+;;; checksum to catch the vast majority of simple mistakes in typing; and with a
+;;; "version" (or "network") prefix that helps identify which use it is for (and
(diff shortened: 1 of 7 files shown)


dbmcclain:
This code represents a massive duplication of effort on existing code already provided by the Crypto system. In Crypto/vec-repr.lisp you will find a unified approach to byte vectors, including BASE58 and HEX string production.

In that module you will find that unsigned byte-8 vectors are just an ordered collection of bytes, accessible by index. They are neither big-endian, nor little-endian. However, by encapsulating with a type structure of one slot, you can declare them to be big-endian, or little-endian, with easy conversions between those two forms, and for extracting the raw vector.

Vectors carry a notion of length. Base58 and Hex, on the other hand are merely integer encodings that carry no implicit notion of length and so are not wholly interchangeable with BEV and LEV. But any BEV or LEV can be stated as a Base58 or Hex value if you like. There will be no forced zero padding on the implied integer values.

markhdavid:
I am sorry, I had not fully realized the base58 support was there so extensively. Of course, I will withdraw my base58 encoder. However, I have an issue with the case where a byte vector starts with a 0: the leading zero seems to get removed.  This seems to be a bug in the base58 encoding. I sent you a note about this on Twist.

markhdavid:
OK talking to David, the base58 encoder encodes a number so leading 0's are just ignored.  I'll work with David to have it do what we need. For now closing this PR while this gets worked out.

PR #170, opened 2018-05-27
Fix use of wordlists

easye:
Someone changed `hash:hash/256` to return a value incompatible with that expected by the dictionary wordlist functions.   

This patch once again ensures that:
```
(let ((i (random (expt 2 255))))
  (= i
     (cosi-keying:convert-wordlist-to-int
      (cosi-keying:convert-int-to-wordlist i))))
```

### DIFF
--- a/src/Cosi-BLS/cosi-keying.lisp
+++ b/src/Cosi-BLS/cosi-keying.lisp
@@ -194,7 +194,7 @@ THE SOFTWARE.
   (check-type val (integer 0))
   (assert (<= (integer-length val) 256))
   (let* ((h  (hash:hash/256 (vec-repr:levn val 32)))
-         (v  (dpb (aref h 0) (byte 8 256) val)))
+         (v  (dpb (aref (vec-repr:lev-vec h) 0) (byte 8 256) val)))
     (loop for ct from 0 below 24
           for pos from 0 by 11
           collect (aref wref (ldb (byte 11 pos) v)))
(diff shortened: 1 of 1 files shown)


dbmcclain (commented):
I'm a bit concerned about the initial ASSERT for integer length <= 256 bits, for something being fed to a hash function. Our public keys are in fact 920 bits at the moment, and subject to change.

Also, be advised that HASH normally returns BEV vectors, so asking for the LEV-VEC of a BEV forces the system to reverse the vector representation of the hash. Is that what you want?

dbmcclain:
Hash normally doesn't care at all how long its input byte source is. The 256 of hash/256 refers to the output length of the hash

dbmcclain:
sorry... the code fragment that Github showed was too abbreviated to make sense of the function code. Looking a the full source, I don't know if I wrote that originally, or if it has been hacked along the way.

But I do see that the intent is to prepend 8 bits from the SHA-3/256 hash onto the little-endian vector representation of the original integer argument, which is expected to be a 256-bit integer. Not sure that actually applies to our Emotiq situation. But the purpose of the 8 bits prepended is to have an integer value some multiple of 11 bits long. 

So, if what you did for yourself works, then okay. But still not sure this even applies to Emotiq.

dbmcclain:
and you are correct that accessing a hash output with AREF is incorrect. HASH returns a typed vector - meaning a vector of bytes encapsulated with a typed structure with one slot. You need to take the vector from that typed vector using either BEV-VEC or LEV-VEC, before accessing with AREF.

So this  looks like very old code, to start with.

PR #171, opened 2018-05-27
157484804 create rest interface wallet

easye:
Initial version of websocket interface for Electron UI delivered under LispWorks

See the commit message of <https://github.com/emotiq/emotiq/commit/0d857633db9edf62a1b8718a87d1ba5909f0d7a4> for description of websocket messages implemented.  

Note that this includes the fix to wordlist mappings to integers covered in <https://github.com/emotiq/emotiq/pull/170>.


src/wallet.lisp
```diff
@@ -18,39 +16,133 @@
...
+   (format nil "wallet/~a/~a"
+           (wallet-name-to-pathname-name name)
+           "emotiq.wallet")
+   (emotiq/user/root)))
    
 (defun wallet-serialize (wallet &key (path (emotiq-wallet-path)))
+  "Serialize WALLET object to PATH"
   (ensure-directories-exist path)
   (with-open-file
       (o path :direction :output :element-type '(unsigned-byte 8))
```
markhdavid, line 107: - There are references to unexported symbols (using double colons ::), e.g., in hexify method. Please eliminate (or explain?)
- What's the magic 31 value in key-phrase? Can the be a comment at least, or a variable or constant naming it?
- there's no :if-exists handler in the with-open-file in wallet-serialize. what happens if the file exists? some lisps and/or os's could cause an error. if the intent is to supersede, should make that explicit.  Or can one backup be kept?  EIther way a staged supersede is probably best: a backup should be kept at least until the writing is done, then it could be gotten rid of.  If the writing is interrupted in the middle, you should not be left with nothing.

markhdavid (changes requested):
A lot of interesting code, a lot of it looks pretty sound and good. Some of the new external modules look interesting and useful, too, but I had not heard of them. 

However, I could not see a description of the project or requirements in the pivotal task. This is a large complex commit with many things going on, including bringing in new external modules.  This needs further explanation and review.  Is a writeup available somewhere -- it could well be that I am just unaware, so if so, can you help point me at it?

One low level objection: several references to non-exported symbols (using double colons ::). Those should not be necessary. I could not approve this commit with those in there even if I did understand enough about all the other stuff.  So consider those at least for now to be minimal requested changes. (There are a few minor notes inline in the "files changed" view. I cannot say I'm near done reviewing, however. If you want to wait for my review to be complete, I understand that, but I need more time on that.)

Thanks.

PR #172, opened 2018-05-28
Remove all top-level uses of PBC:MAKE-KEY-PAIR

easye:
Such usage results in segmentation violations at production runtime,
as top-level references during LispWorks delivery process refer to
malloc()'d memory.


guitarvydas (approved):
Appears to work on linux/mac.  Creates binaries on both.  Fails in Travis due to known problem (OUTPUTTER0).

PR #173, opened 2018-05-29
157840690 pt mte fix binary delivery c

guitarvydas:
This should fix the Travis build problem, allowing :gossip-tests to run multiprocessing under Travis.  There are many comments in emotiq/etc/deliver/delivery.lisp, emotiq/src/Actors/actors-startup.lisp and emotiq/src/startup.lisp.

I've chosen to put the "special cases" into the binary workflow.  The binary MUST call EMOTIQ:START as its entry point.

Actors-startup.lisp allows all non-Lispworks users to simply LOAD the code (which automatically starts up Actors) in "development" mode.  

The #+Lispworks case requires 3 possible choices (done without adding to *features*).  (1) David mode - uses the LW-only DEFINE-ACTION function, (2) nothing, for binary building, can't be running multiprocessing when calling DELIVER and (3) all other "development" mode use-cases - just load and startup Actors without doing anything special.  I set up a special var in emotiq/deliver/deliver.lisp and remove it in EMOTIQ:START for the binary-building workflow.

### DIFF
--- a/src/startup.lisp
+++ b/src/startup.lisp
@@ -2,6 +2,7 @@
 
 (defun main (&optional how-started-message?)
   (message-running-state how-started-message?)
+  (actors:install-actor-system)
   (pbc:init-pairing)
   (let ((context (start-blockchain-context)))
     (with-blockchain-context (context)
--- a/src/startup.lisp
+++ b/src/startup.lisp
@@ -28,8 +28,7 @@
   (message-running-state how-started-message?)
   (format *standard-output* "Making key pair…")
   (let ((keypair (pbc:make-key-pair :foo)))
-    (format *standard-output* "  Created ~a~&" keypair))
-  #+lispworks(lispworks:quit))
+    (format *standard-output* "  Created ~a~&" keypair)))
 
 ;; Entry Point for binary version of the system.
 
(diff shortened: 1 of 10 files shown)


easye:
@guitarvydas Does this fix the running of `gossip-tests`?  

If so shouldn't we re-enable the tests as disabled via <https://github.com/emotiq/emotiq/commit/0b5b9c65a5dcd32be659459a2762e56dad851c73>

src/startup.lisp
```diff
@@ -1,32 +1,61 @@
...
-    (format *standard-output* "  Created ~a~&" keypair)))
-
-(defun message-running-state (&optional how-started-message?)
-  (format *standard-output* "~%Running ~a in ~a~%with args [~a]~%"
-          (or how-started-message? "interactively")
-          (if (production-p) "production" "development")
-	  (argv)))
-
+    (format *standard-output* "  Created ~a~&" keypair))
+  #+lispworks(lispworks:quit))
```
easye: @guitarvydas 

The semantics of `main` are to invoke the application at development and production time.

Causing lispworks to unconditionally quit breaks the ability of a developer to work on this function.  

If you really wish to unceremoniously quit LispWorks, you should add this to the end of `start` which is documented as only being called at production runtime.

easye (changes requested):

easye (approved):

PR #175, opened 2018-05-29
socket-actors.lisp: Switch to unified "select loop" [not debugged]

lukego:
*NOTE: This code is not properly tested yet and the PR is created for feedback in the meantime.*

Switch socket-actors to using a single process to poll for data from all peers. (This replaces the previous model of polling each connection with a separate thread.)

The process waits on all open connections using:

  (USOCKET:WAIT-FOR-INPUT :READY-ONLY T :TIMEOUT 0.01)

The timeout means that the select is restarted at regular intervals to pick up newly opened connections.

This code is not yet debugged!


lukego:
Closing this PR since it is in a preliminary state and we are discussing this code effectively on Twist.

PR #176, opened 2018-05-29
Dbm 0529a

dbmcclain:
Fix mempool and utx-table databases to become updated only at end of successful COMMIT phase of consensus. Meanwhile, a parallel transaction changes table is maintained for use during PREPARE phase validation.

At end of commit phase, replay transactions from successful block, regardless of witness opinion during prepare phase.

Remove undefined symbols from cosi-handlers.lisp (was calling emotiq/sim:check...)

Short circuit the UDP protocol for all nodes on one host Lisp image. Allows full consensus protocol checking without bypassing consensus threshold checking.


src/Cosi-BLS/config/cosi-keying.txt
```diff
@@ -0,0 +1,19 @@
+
```
easye: @dbmcclain The files under `src/Cosi-BLS/config/` are not supposed to be under git revision control.  Or has this changed for some reason?  

Either remove, or explain why these files should be part of git (and adjust the top-level `.gitignore` patterns.

dbmcclain: Hmm... must be a leftover from testing with Paul. I will remove and resubmit.

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -872,49 +986,40 @@ check that each TXIN and TXOUT is mathematically sound."
     (and (check-byz-threshold bits blk) ;; check witness count for BFT threshold
          (check-block-transactions-hash blk)
          ;; check multisig as valid signature on header
-         (pbc:check-hash hash sig mpkey)
-         ;; ensure multisign pkey is sum of witness pkeys
-         (int= mpkey wsum))))
+         (pbc:check-hash hash sig (clos:change-class wsum 'pbc:public-key)))
```
easye: @dbmcclain Remove unneeded `clos` prefix.

markhdavid (approved):
reviewed several issues and resolved several points with Mark E. and David together. looks good. approved.

PR #177, opened 2018-05-29
Monads

svspire:
Some changes to make gossip-startup work in parallel and be more reliable in the face of remote machines that are not listening. Results and errors are now returned and coalesced as metadata-containing monads.


svspire:
Note: Running this with
`(asdf:make :gossip)`
fails because `*gossip-db-file*` etc are undefined.
But 
`(ql:quickload :gossip)`
works fine. Why?

PR #178, opened 2018-05-29
fix symbol typo

dbmcclain:


easye (approved):
That's what we get for approving changes without testing them.  

Obviously approved, and will merge myself.

PR #180, opened 2018-05-29
bitcoin-style base58-check p2pkh addresses

markhdavid:
bitcoin-style base58-check p2pkh addresses

- vec-repr:base58 method now left-pads with encoded 0's in vector case
- added base58 encoding unit tests (lifted from bitcoin's)
  test with: (asdf:test-system :crypto-pairings/t)
  notice: Base58-test
- added new hashes: hash/ripemd/160 and has/sha2/256
- added new module address.lisp which exports public-key-to-address
- created new unit tests file address-tests.lisp
  test with (asdf:test-system :cosi-bls-tests)
  notice: "Public Key to Key Hash as Address Test (Base58Check Encoding)"

### DIFF
--- a/src/Cosi-BLS/address.lisp
+++ b/src/Cosi-BLS/address.lisp
@@ -1,0 +1,152 @@
+;;;; address.lisp
+
+(in-package :cosi/proofs)
+
+
+
+;;;; Addresses: Public Key Hashes in Emotiq
+
+;;; Hashed public keys are presented as `addresses'.  An `address' for a public
+;;; key is in general produced as follows:
+;;;
+;;; First, it is hashed irreversibly, and then encoded in a user friendly
+;;; manner, that is, shorter; with characters easily typed, read, and
+;;; distinguishable by humans in nearly all computing environments; with a
+;;; checksum to catch the vast majority of simple mistakes in typing; and with a
+;;; "version" (or "network") prefix that helps identify which use it is for (and
+;;; also which cryptocurrency it's for (e.g., where Bitcoin uses 0x00, Litecoin
+;;; uses 0x30).
+
+;;; As many other cryptocurrencies have done, we simply adopt the Bitcoin
+;;; specification for addresses, known as Base58Check Encoding. The only thing
+;;; that we makes our addresses distinct is that we use unique version/network
+;;; prefix. (We hope ours is unique. There does not seem to be any central
+;;; registry.  If there is, we do not know of one at this time, May 27, 2018.)
+
+;;; So for the Bitcoin address the details are as follows. First we hash the
+;;; public key with SHA2/256. The resulting hash is shortened with a second hash
+;;; of RIPEMD160, getting bits down to 160 or 20 bytes (octets).  The version
+;;; prefix octet is prepended on to the front of the resulting hash result,
+;;; viewed as an octet vector.  The resulting vector V gets a 4-element checksum
+;;; vector appended containing elements created as follows: take a double
+;;; sha2/256 hash of V, and take the first 4 elements of the result viewed as an
+;;; octet vector, and then append that octet vector at the end of V. Finally,
+;;; that entire result is base58-encoded as a string, yielding the final result.
+;;;
+;;; Here is a more concise summary:
+;;;
+;;;   Public Key => SHA-2/256 => RIPEMD160 => Base58Check Encode version prefix
+
+;;; The byte (octet) layout ends up as follows:
+;;;
+;;;   [version] [RIPEMD160 hash] [checksum]
+;;;       1           20              4
+
+
+;;; The constants +mainnet-version-for-public-key-to-address+ and
+;;; +testnet-version-for-public-key-to-address+ are to be used for mainnet and
+;;; testnet deployments, respectively, as the version (network) prefix octet.
+
+;;; The values chosen are somewhat arbitrary but with the following constraints:
+;;;
+;;;   (1) cannot conflict with prominent well-known Bitcoin values, notably
+;;;
+;;;         0x00 0x03 0x6F 0x80 0x0142 0x6F 0x00;
+;;;
+;;;   (2) ideally, ought not to conflict other prominant cryptocurrencies'
+;;;   established mainnet p2pkh version prefixes, notably
+;;;
+;;;     Lightcoin: 176; Doge: 30; Dash: 76
+;;;
+;;;   (3) initial character of resulting address string must be unique between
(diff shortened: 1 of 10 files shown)


dbmcclain:
Two issues here...

1. What is the purpose of including SHA2/256? It is an outmoded hash function and current NIST recommendations are to avoid it in new code.

2. I see the prepending of zeros on BASE58, but I don't see any way to tell it how long the result should be, and when to use that padding.

Third issue, when I check out this branch, and perform:
(in-package :pbc)
(setf k (make-key-pair :dave))
(setf pkey (keying-triple-pkey k))
(base58 *) ;; --> goes into infinite loop

The correct result, from dev branch, is:

#<BASE58 8jUuBXBRfFD8edN6a6AVSiQ7cpgXdw9334hoKCuyF8euTM2vt8HaegspU9i7M8qfZGn6ob3Z1gsYFeK6bNfTmqHW25WXoyLfot5s78rzb755ARgg23KhSDBUBmwXP7tALUV2S28erhD9CyCLkg58gKD9kGrL >

dbmcclain:
I think one source of confusion is that BASE58 is an integer encoding, not a vector encoding. An augmented form of BASE58 encoding could be defined so that it includes a vector length represented by the base58 encoding. But that would be a composite form. 

I actually had such a form at one time, and decided that neither HEX nor BASE58 were really vector encodings at all. 

And while you certainly can convert a vector form to base58 and hex strings, even with leading 0 bytes in the big endian representation, you might not get that same vector back if you convert to base58 then back again with BEV.  But a side function BEVN can be told how long the conversion should be, and does so with prepended zero bytes when the desired result is longer than the natural conversion back to vector form.

HEX actually could become a vector conversion directly, since it is based on power-of-two base. But BASE58 never could be. There is no grouping of bits in a vector that can give rise to base58 encoding. It is, and will always be, an integer encoding.

markhdavid:
Answering the 2 issues:  

1. Use of sha2/256 is for the purpose of implementing this encoding. We are primarily doing it to be compatible with a convenient standard for addresses. This makes for convenient testing, allows for some interoperability with wallets, and will look professional and familiar to users of other cryptocurrencies, as has already been discussed.  Any deficiencies or concerns with sha2/256 on its own are dealth with by having the additional second hash (ripemd160), as was previously also discussed.

2. the length of the prepending is determined by the number of 0 bytes in at the beginning of the vector.

3. I don't know what's causing this issue. I'll track this down now. Thanks for bringing this to my attention.

dbmcclain:
"the length of the prepending is determined by the number of 0 bytes in at the beginning of the vector."

Well, that can never be technically correct because 58 is not a power of 2.

markhdavid:
Regarding integer encoding vs. vector encoding, we simply need a vector encoding.  There's no known use in our software for the integer encoding, nor for the decoding, but we have (and the world has) the vector encoding case very strongly, and it's being used about 2 million times a day for bitcoin address generation, so if you want, we can just use my simple function I wrote 2 days ago to generate it from a vector if you want to leave your code pure.  This is bumming me out.

dbmcclain:
The issue revolves around the fact that 58 is not a power of 2. No grouping of bits can ever be used for convert from base58 representation to a vector, and vice versa. Hence conversion to integer form is always needed. So the next question becomes, is the vector to be big-endian, or little-endian?

dbmcclain:
Honestly, I would have expected the function to generate addresses to require no more than a dozen lines of code, probably fewer.

markhdavid:
I think there are about 20 odd lines of code in address.lisp. Let's not worry about that.  I'll take another shot at fixing the vec-repr.lisp stuff, unless you want to take over.

dbmcclain:
I suggest that if you want a full-length conversion to base58, then be sure that the MSB of the integer equivalent of the vector be set to 1. Only then can you be sure to have a full-length base58 encoding. Perhaps that is what Bitcoin does?

markhdavid:
I am sorry, I really did not see how it could possibly work to set the MSB of the integer to 1.    I do not know of anything like that done in Bitcoin code for doing base58 encoding.  Possibly I'm not following this well enough.

I fixed the bug I had introduced when I transformed the code in vec-repr: in the non-vector case, it needed to apply (INT X).  Should have just broken out a separate method based on arg type, so that's what I've done now. 

Also made testing a bit more verbose. (I've seen some cases where there is one failure indicated, but then it goes away. I'm not sure where the failure is, since verbose is off, other than it is somewhere in crypto tests. Kind of a "heisenbug".

markhdavid:
The CI was stalled:

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

dbmcclain:
I seem to recall that Bitcoin likes to use prefix bytes to indicate some sort of type. Those prefix bytes contain bits that are prepended to the key vector, ahead of conversion to base58. So setting the MSB is done in that manner.

markhdavid:
Bitcoin’s mainnet P2PKH prefix is actually 0x00. 

Bitcoin’s test suite for encodings, which I’ve incorporated into unit tests I created, include a hex string starting “00...”, where its expected output begins “1...” — exactly the leading zero case.

dbmcclain:
Okay, while I continue to disagree that the number of leading zero bytes in the vector has anything directly to do with the required number of padding chars on a base58 representation, the code is a credible performer, and I will go ahead and approve. The future will dictate whether this is a proper encoding system.

I just merged this branch with dev branch, and a typo correction in the networking layer, from my branch, and pushed back. 

This code continues to run the emotiq/sim tests properly for both uncloaked and cloaked transactions. It also properly withstands repeated (emotiq/sim::initialize).

Note that while leading zero bytes in a vector -> base58 get represented as "1", the inverse conversion produces a BEV without leading zero bytes. So the conversion is not directly reversible. But then it never was to begin with either.

dbmcclain (approved):

dbmcclain:
I don't understand the Travis failures. But as far as I can tell, it has nothing to do with any of the code committed in this pull request. Looks like the failures were inherited from the merge with dev branch.

As for encodings, it all looks to me like "Paint-by-Numbers" programming - we do it this way because the other guys do it this way too. Nothing sensible justifies it. But oh well...

markhdavid:
OH WOW, just noticed this: I had read where David wrote about approving and merging, and I'd assumed he'd merged it into dev and closed this, but I see he's been committing to this branch and merging it in with his branch.  I do not understand why this was done. I'll continue reviewing.

markhdavid:
Although it was not the greatest procedure, in the interest of expediency I decided to let this go in via David's PR #184, which incorporates this PR. Accordingly, I'm closing this PR.

PR #181, opened 2018-05-29
Sim binary

guitarvydas:
This runs node-sim as a binary.  Merged with dev (dbm 0529a).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,12 @@ script:
              :gossip-tests";
       systems=${systems} bash etc/test-harness.bash
 
+# was:  (removed/commented out :gossip-tests pending fix to actors initialization issue.  -mhd, 5/28/18)
+#  - systems=":cosi-bls-tests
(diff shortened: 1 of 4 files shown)


easye (changes requested):
This is the wrong order to declare dependencies.  

`deliver.lisp` should only Quickload `emotiq/startup`.  `emotiq/deliver` should only depend on `emotiq` to get the `emotiq` package declaration.

`emotiq/startup` should declare the dependencies that it needs to run, which would include `emotiq/sim`.

`emotiq/sim` should not depend on `emotiq/startup` as it has no dependencies on the main loop.

.travis.yml
```diff
@@ -60,6 +60,12 @@ script:
              :gossip-tests";
       systems=${systems} bash etc/test-harness.bash
```
easye, line 62: Remove this commented out block:  we're using revision control.

PR #182, opened 2018-05-30
157661420 migrate ci pipeline 20180530a

echupriyanov:
GitLab CI pipeline


etc/build-binary.bash
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+set -x
+
+KIND=${1:-production}
+DIR="$(cd -P "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+BASE=${DIR}/..
+
+if [ ${var} ]
```
easye: I think this is wrong.  `var` is not defined until line 29, so unless `var` is passed from the environment, this test always fails.

easye (commented):
Is there a use of `run-emotiq.bash` that works outside of the the CI pipeline?  Locating this script in `etc/` rather than `ci/` is perhaps slightly confusing to the developer.

easye (commented):
The invocation of the `etc/build-emotiq.bash` script still references the LispWorks implementation stored under a Roswell pathname by default, but I suppose that is because the LispWorks installation process installs in the Roswell hierearchy.  Not necessarily a problem, just noting that this would probably confuse a person who doesn't know the history of how our tools were developed.

easye (commented):
It might be nice to have a README in the etc directory, addressed to a developer explaining what various scripts are used for.

easye (approved):
All my previous comments are just suggestions, not reasons to block this pull request.  

Developers can continue to build binaries via `etc/build-emotiq.bash`.  New CI runs fine.

Approved.

echupriyanov:
Great thanks for feedback. Will update now.

etc/build-binary.bash
```diff
@@ -30,6 +30,11 @@ etcdeliver=${etc}/deliver
 prefix=${var}/local
 lib=${prefix}/lib
 
+if [ ${var} ]
```
easye: I still think you mean 

```
if [[ -d ${var} ]]
```
as your test just tests whether `var` evaluates to a non-zero length string.

easye (approved):
Thanks for the adjustment to the README.  

Let's just approve this…

etc/build-binary.bash
echupriyanov: Oh, of course. That peace is from Paul's `production.bash`, and I didn't take a closer look.

PR #183, opened 2018-05-30
addressed comments from pr181 ; moved start in front of main …

guitarvydas:
…; moved simulation calls to start, removed from main ; addressed issues from PR181

### DIFF
--- a/src/startup.lisp
+++ b/src/startup.lisp
@@ -23,13 +23,6 @@
 
 (in-package "EMOTIQ")
 
-;; "Entry Point" for development - does nothing, just load and go
-(defun main (&optional how-started-message?)
-  (message-running-state how-started-message?)
-  (format *standard-output* "Making key pair…")
-  (let ((keypair (pbc:make-key-pair :foo)))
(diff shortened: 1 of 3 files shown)


src/startup.lisp
```diff
@@ -42,7 +35,17 @@
   (setq *production* t)  ;; used by EMOTIQ:PRODUCTION-P in Crypto
   (message-running-state "from command line")
   (actors:install-actor-system)
-  (main))
+  (main)
+  (emotiq/sim:initialize)
```
easye, line 39: You need to move these functions into the main loop.

src/startup.lisp
```diff
@@ -42,7 +35,17 @@
...
+  (emotiq/sim:run :cloaked nil))
+
+;; "Entry Point" for development - does nothing, just load and go
```
easye, line 42: This comment is incorrect:  it isn't true that this function does nothing.  In fact it is meant to be the main loop of the application.  

Better to do as a docstring in any event.

easye (changes requested):

PR #184, opened 2018-05-31
Dbm 0531a

dbmcclain:
Simplify Cosi protocol. No longer passing entire block back to leader from witness nodes. Just have witnesses return their signature and bitmap bits. 

Remove Epoch slot from Block representation.

Unfortunately, without having a block structure to refer to for the formation of composite public keys, the notary service in the old Cosi simulator had to be removed. We no longer use that, as it only tested a single pass through the Cosi network. That removes cosi-simgen::tst.


markhdavid (changes requested):
David, I have no idea why you decided to incorporate my bit "bitcoin-style base58-check p2pkh addresses" PR into this PR.  I'd assumed you'd merged that into dev.  I would like to make that commit be a single squashed commit on dev, and then you can pull it into your branch, and do these commits on top of that.  Or alternatively do these commits first and then I can do my commit afterwards.  I do not want to make more work for you, but I do not want my PR "bitcoin-style base58-check p2pkh addresses" combined with this PR.

dbmcclain:
Well, Mark, I merged with dev branch just prior to making this pull request. So it would seem that what you want has already been accomplished.

But you very likely know a whole lot more about Git than I do. So do what you think needs doing.

markhdavid:
Oh well, it's not that important. I should have paid better attention at the time. What I would like to do now would leave you a mess that would require even more git craft knowledge than average, so I'll just go with your approach.  I built it and ran it in my environment. Compiles as clean as before, and tests run as well as before. So, approved.

markhdavid (approved):

markhdavid:
My earlier PR #180 will have to be closed now in favor of this, since this incorporates it. I'll do that now.

markhdavid:
I'll now go ahead and push the "Rebase and merge" button to close out this PR.

PR #186, opened 2018-06-01
Define defstub - used to still compiler re many unimplemented functions.

markhdavid:
This commit is primarily intended to eliminate pesky warnings.

For each function X in the crypto library that was not defined that we
got complaints about being called but not defined, there's now a
(defstub X) form.  This makes sure it's defined, although calling it
would not do anything really useful.  It also makes it known that the
function is a stub by adding a a property to the function name. That
property gets used by the little WITH-FAST-IMPL facility to avoid
generating a call to it.  I also cleaned up the error handling in
WITH-FAST-IMPL: previously, it would silently trap any error, and then
call makeunbound.  That's not a very clean way to handle it. First, it
should at least print the error, and help keep track of it.  Second,
it should not use makunbound. Instead, it now puts a flag property on
the function name and uses that as the gate to decide whether to call
the "fast" function.

NB: the speed of checking fboundp vs. property checks should be about
the same or noise level difference in most Lisps.  However, I suspect
either check might be adding significant overhad to the raw "fast"
call; review at a later date when we do optimization.

### DIFF
--- a/src/Crypto/ecc-B571.lisp
+++ b/src/Crypto/ecc-B571.lisp
@@ -195,6 +195,9 @@ THE SOFTWARE.
       :alpha 1))
     ))
 
+
+(defstub c-ecc571-add)
+
 (defun ecc-double (pt)
   (optima:match pt
     ((ecc-infinite-)  pt)
@@ -273,6 +276,10 @@ THE SOFTWARE.
                   (lambda (pt)
                     (ecc-projective pt :alpha alpha)) ))
  
+
+
+(defstub c-ecc571-mul)
+
 (defun ecc-affine-mul (pt n)
   ;; left-to-right algorithm
   ;; for affine coordinates
--- a/src/Crypto/gf-571.lisp
+++ b/src/Crypto/gf-571.lisp
@@ -324,6 +324,10 @@ THE SOFTWARE.
                 (setf ans (logxor ans a)))) )
         ))
 
+
+(defstub c-gf571-mul)
+(defstub c-gf128-mul)
+
 (defun bin-gf* (a b)
   (cond
    ((or (zerop a)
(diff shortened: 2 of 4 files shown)


markhdavid:
Previously, warnings had been as follows
```

The following functions are undefined:
ECC-CRYPTO-B571::C-GF571-INV which is referenced by ECC-CRYPTO-B571::GFINV
ECC-CRYPTO-B571::C-GF128-INV which is referenced by ECC-CRYPTO-B571::GFINV
ECC-CRYPTO-B571::GF-RANDOM-K* which is referenced by ECC-CRYPTO-B571::ECC-FIND-A-VALID-X, ECC-CRYPTO-B571::ECC-MUL, and ECC-CRYPTO-B571::MAKE-ECC-PROJECTIVE-PT
ECC-CRYPTO-B571::C-ECC571-ADD which is referenced by ECC-CRYPTO-B571::ECC-ADD and ECC-CRYPTO-B571::ECC-DOUBLE
ECC-CRYPTO-B571::C-KDF which is referenced by ECC-CRYPTO-B571::APPLY-C-KDF
ECC-CRYPTO-B571::C-GF571-DIV which is referenced by ECC-CRYPTO-B571::BIN-GF/
ECC-CRYPTO-B571::C-GF128-DIV which is referenced by ECC-CRYPTO-B571::BIN-GF/
ECC-CRYPTO-B571::SHA2_FILE which is referenced by ECC-CRYPTO-B571::FAST-SHA2-FILE
ECC-CRYPTO-B571::SHAD2_FILE which is referenced by ECC-CRYPTO-B571::FAST-SHAD2-FILE
ECC-CRYPTO-B571::C-GF571-MUL which is referenced by ECC-CRYPTO-B571::BIN-GF*
ECC-CRYPTO-B571::C-GF128-MUL which is referenced by ECC-CRYPTO-B571::BIN-GF*
ECC-CRYPTO-B571::C-ECC571-MUL which is referenced by ECC-CRYPTO-B571::ECC-BASIC-MUL

```

dbmcclain (approved):
I never see these warnings in my LWM system. Does this happen on some other configuration/Lisp?

At any rate, none of these functions are used by us, and so having stubs to snuff warnings is fine.

PR #187, opened 2018-06-01
fairly obvious code changes, getting rid of warnings

markhdavid:
wallet-deserialize now looks much more like wallet-serialize above it in handling path defaulting, thereby losing warning

create-wallet made no use of force-p (to check if arg supplied), so got rid of it, thereby losing warning

### DIFF
--- a/src/wallet.lisp
+++ b/src/wallet.lisp
@@ -35,16 +35,15 @@
     (lisp-object-encoder:serialize wallet o)))
     
 (defun wallet-deserialize (&key (path (emotiq-wallet-path)))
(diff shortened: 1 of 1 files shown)


markhdavid:
Previous warnings compiling src/wallet.lisp in LispWorks:
```
;;;*** Warning in EMOTIQ/WALLET:WALLET-DESERIALIZE: PATH is bound but not referenced
;;;*** Warning in EMOTIQ/WALLET:CREATE-WALLET: FORCE-P is bound but not referenced
```

easye (approved):

PR #188, opened 2018-06-01
Fix warning re *current-node*, move cosi-netw-xlat after cosi-handlers

markhdavid:
Moved module that used *current-node* to later in load order after its defvar is loaded.

### DIFF
--- a/src/Cosi-BLS/cosi-bls.asd
+++ b/src/Cosi-BLS/cosi-bls.asd
@@ -50,10 +50,10 @@ THE SOFTWARE.
                                      (:file "block")
                                      (:file "cosi-keying")
 				     (:file "cosi-construction")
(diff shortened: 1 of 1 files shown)


markhdavid:
Previous warnings given by LispWorks compiling src/Cosi-BLS/cosi-netw-xlat.lisp:
```
;;;*** Warning in (METHOD ACTORS:SEND (PBC-INTERFACE:PUBLIC-KEY)): COSI-SIMGEN:*CURRENT-NODE* assumed special
;;;*** Warning in (METHOD ACTORS:SEND (COSI-SIMGEN:NODE)): COSI-SIMGEN:*CURRENT-NODE* assumed special
```

dbmcclain (approved):
whatever works... looks like a benign change to semantics. Approved

PR #189, opened 2018-06-01
Dbm 0601a

dbmcclain:
Fix SBCL compatibility issues with Crypto and Actors. Modified SDLE-STORE to use original local definitions for DELAY/FORCE instead of UM.LAZY:LAZY/FORCE. Remove futures.lisp from actors.asd,

Extend definition of Actors RECV to define a lexical function RETRY-RECV visible only within RECV body. That eliminates the need to plant RECV blocks inside LABELS for recurrent use of the RECV.

Clean up Cosi and Elections code based on the now simplified syntax permitted with RETRY-RECV.


src/Actors/actors.asd
```diff
@@ -35,7 +35,7 @@ SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
                 ;; (:file "actors-mb")
                 (:file "actors-machines")
                 (:file "actors-startup")
-                (:file "futures")
+                #+removed (:file "futures")
```
easye: Please don't define new features like "removed".  If you wish to remove forms via a null conditional use something like 
```
#+(or)
```
which will always be false. 

But we don't we just remove the reference to futures, along with the "futures.lisp" conditional?

easye (commented):
This request doesn't address the problem that I was trying to address for `core-crypto` (and its dependency`sdle-store`) in depending on `actors`.  With this patch, `sdle-store` still depends on `actors`.  

One can observe the dependency chain needed by via loading a given system by the following code:

```
(defvar cl-user::*asdf-systems-loaded* nil)
(defmethod asdf:perform :before ((operation asdf:load-op) (system asdf:system))
  (push system cl-user::*asdf-systems-loaded*))
```


I tried using the version of `lazy` actually defined in `useful-macros`, but that version contains LispWorks specific use of functions in the `sys` and `hcl` packages mainly to use thread-safe memory access.  

```
diff -r 210b256ffa7c src/sdle-store/sdle-store.asd
--- a/src/sdle-store/sdle-store.asd	Fri Jun 01 10:28:46 2018 -0700
+++ b/src/sdle-store/sdle-store.asd	Sat Jun 02 07:59:52 2018 +0200
@@ -49,8 +49,7 @@
   :long-description "Portable CL Package to serialize data"
   :licence "MIT"
   :in-order-to ((test-op (test-op "sdle-store/tests")))
-  :depends-on (useful-macros
-               actors) ;; Actors contains the improved defintion of UM.LAZY
+  :depends-on (useful-macros)
   :serial t
   :components ((:file "package")
                (:file "utils")
diff -r 210b256ffa7c src/useful-macros/useful-macros.asd
--- a/src/useful-macros/useful-macros.asd	Fri Jun 01 10:28:46 2018 -0700
+++ b/src/useful-macros/useful-macros.asd	Sat Jun 02 07:59:52 2018 +0200
@@ -55,7 +55,7 @@
                 (:file "critical-section")
                 ;; (:file "dispatch-queues") ;; what do we need these for?
                 
-                ;; (:file "lazy") ;; not supplanted by Actors
+                (:file "lazy") 
                 
                 #+(AND :COM.RAL :LISPWORKS) (:file "remembered-filenames")
                 ;; (:file "useful-macros-old")
```

easye (approved):
Despite not addressing the request to remove the need to load the `actor` system for `core-crypto`, this code does not make the current situation worse, so if @dbmcclain  is happy with these changes, I approve the merge into `dev`.

dbmcclain:
I found two more references to um.lazy in sdle-store/default-backends.lisp. Sorry for missing those before. Perhaps this now fixes the dependency on Actors?

PR #190, opened 2018-06-02
Fixes for compilation under SBCL

easye:
With these changes, `emotiq/sim` will compile under SBCL.  

The blockchain simulation doesn't seem to complete correctly, but at least things compile

### DIFF
--- a/src/Actors/actors.asd
+++ b/src/Actors/actors.asd
@@ -29,8 +29,8 @@ SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
   :author      "D.McClain <dbm@refined-audiometrics.com>"
   :license     "Copyright (c) 2017 by Refined Audiometrics Laboratory, LLC. MIT License terms apply."
   :components  ((:file "packages")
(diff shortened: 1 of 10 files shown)


dbmcclain:
I checked out this branch and examined the code, in particular in modular-math.lisp. It appears that SBCL has trouble with the optional 3rd argument in DOLIST? In detail the code looks correct as rewritten.

I loaded up emotiq/sim just fine (no apparent errors) and ran it in cloaked mode to double check the performance of modular-math changes. Bulletproofs exercise the modular-math package by a large amount. Everything looks okay.

Approved

dbmcclain (approved):

easye:
@dbmcclain 
>I checked out this branch and examined the code, in particular in modular-math.lisp. It appears that SBCL has trouble with the optional 3rd argument in DOLIST? […]

Yes, SBCL had enough trouble analyzing the type of the `dolist` result form that the compilation would fail.  I think the problem was in the compiler failing to determine the return value of a function that ends with such a `dolist` but I didn't spend that much time investigating as the explicit rewrite was simple enough.

PR #191, opened 2018-06-04
sdle-store: explicitly remove unneeded actors system dependency

easye:
Finish what we started:  unless we remove the ASDF dependency of `sdle-store` on `actors`, loading `core-crypto` will still kick off the actors `load-toplevel` initialization routine.  

One can examine what is loaded via ASDF with code like:
```
(in-package :cl-user)

(defvar cl-user::*loaded-asdf-systems* nil)

(defmethod asdf:perform :before ((operation asdf::load-op) (system asdf:system))
  (push system cl-user::*loaded-asdf-systems*))
```


dbmcclain (approved):
Changes appear benign to behavior of systems

PR #192, opened 2018-06-04
158026183 add one retry emotiq ci jobs

echupriyanov:
* Added timeout watchdog for running tests
* Added 1 extra attempt to run failed tests job


easye (approved):
I really think scripts that rely on bash rather than Bourne shell features should end in `bash` rather than `sh`, but this is a minor nit.

Approved.

PR #193, opened 2018-06-04
Dbm 0604a

dbmcclain:
Was using INT= to compare hashes, public keys, signatures, etc. Not really appropriate because conversion to INT ignores leading null bytes of vector quantities. Instead, we should use VEC= which first ensures that each operand is a BEV, and then compares byte by byte of the vectors.

The situation is more subtle than just described. ECC points (compressed form) can be legitimately compared with INT=. Hash values, which can easily have leading zero bytes should be compared with HASH=. Arbitrary raw vectors can be compared with VEC=. 

When hash values are used as Shamir-Fiat values, they can be compared with INT= (The object of Shamir-Fiat challenges is to compute some random challenge value, generally based on the hash of public information in the proof transcript.)

HASH= is a defmethod that expects two HASH class values. In our final public form of the blockchain, it is possible that these items will be stored as raw C vectors, in which case some type conversions will be needed before HASH= is used.

### DIFF
--- a/src/Actors/actors.asd
+++ b/src/Actors/actors.asd
@@ -35,7 +35,7 @@ SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
                 ;; (:file "actors-mb")
                 (:file "actors-machines")
                 (:file "actors-startup")
-                #+(or) (:file "futures")
+                #+(AND :COM.RAL :LISPWORKS) (:file "futures")
                 #+(AND :COM.RAL :LISPWORKS) (:file "linda-tuples")
                 #|
                 (:file "actors-macros")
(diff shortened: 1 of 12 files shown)


easye (commented):
As long  as <https://github.com/emotiq/emotiq/pull/193/commits/c4b68af95ff9d73d5b21a909178c914da8e4c761> is the last commit to this pull request, the changes are fine.

easye (approved):

easye:
In running the tests defined in `crypto-pairings/t` under SBCL
```
(asdf:test-system :crypto-pairings/t)
```
I noticed that SBCL fails on the hash functions comparing strings with the basis consistency section <https://github.com/emotiq/emotiq/blob/dev/src/Crypto/tests/crypto-tests.lisp#L10>.   Both CCL and LWPRO succeed in the tests.

This is not a regression to `dev` (i.e. `dev` fails these tests as well under SBCL), which indicates that something is off in the SBCL code wrt. hashing.

dbmcclain:
makes me wonder if something is off in Ironclad under SBCL...

PR #194, opened 2018-06-05
Added some verbosity to `test-harness` results

echupriyanov:
Just a handy report messages to see actual tests results as returned by `(asdf:test-system system)`

### DIFF
--- a/etc/test-harness.lisp
+++ b/etc/test-harness.lisp
@@ -14,10 +14,10 @@
   (handler-case
       (if (asdf:test-system system)
           (progn
-            (format *standard-output* "~&Test passed for system: `~a`~&" system)
+            (format *standard-output* "~&Test-system succeeded for system: `~a`~&" system)
             (uiop:quit 0))
           (progn
-            (format *standard-output* "~&Test failed for system: `~a`~&" system)
+            (format *standard-output* "~&Test-system failed for system: `~a`~&" system)
             (uiop:quit -1)))
     (error (e)
       (format *standard-output*
--- a/etc/test-harness.lisp
+++ b/etc/test-harness.lisp
@@ -5,7 +5,7 @@
 (in-package :test-harness)
 
 (defun test-system (system)
-  (handler-case 
+  (handler-case
       (ql:quickload system)
     (error (e)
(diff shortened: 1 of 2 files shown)


etc/test-harness.lisp
```diff
@@ -5,16 +5,20 @@
...
     (error (e)
       (format *standard-output*
               "~&Failed to Quickload dependencies for system `~a`:~&~a~&" system e)
       (uiop:quit -1)))
   (handler-case
       (if (asdf:test-system system)
-          (uiop:quit 0)
-          (uiop:quit -1))
+          (progn
+            (format *standard-output* "~&Test passed for system: `~a`~&" system)
```
easye: Strictly speaking, the semantic is that the test system invocation either succeeded or failure, not whether the tests passed.  Individual tests may fail during a `asdf:test-system` invocation while still returning a non-nil value. 

Better wording might be "Test system succeeded for system…" and "Test system failed for system…".

easye (approved):
Invocation works for me.

etc/test-harness.lisp
echupriyanov: Thanks for feedback. What I'm actually trying to solve is to detect situation like happened in this job: https://gitlab.aws.emotiq.ch/emotiq/emotiq/-/jobs/1081
```
[package gossip-tests].
Unit Test Summary
 | 0 assertions total
 | 0 passed
 | 0 failed
 | 3 execution errors
 | 0 missing tests

<== ASDF:TEST-SYSTEM on :gossip-tests succeeded.
```
I'm not sure this is a correct behavior. Or am I missing something?

easye: It is "correct" but undesired.  We need to figure out how to get `lisp-unit` to return a non-zero result to the `asdf` machinery.

easye: From a look through the code for asdf-3.3.1, it unfortunately seems that `asdf:test-system` *always* returns `t` on invocation.  This seems to indicate that we can't get a non-zero status for failing tests running down this route.

echupriyanov: Time to fork `asdf`?

PR #196, opened 2018-06-05
Pt nowatchdog2

guitarvydas:
This commit removes the watchdog timer from Actors.  The code is still there in case we want to re-enable it during testing, using by pushing \:emotiq-watchdog onto \*features\*.

(@MarkE, I know it's not your day for reviewing, but you know the most about the reasoning behind this commit).

### DIFF
--- a/src/emotiq.asd
+++ b/src/emotiq.asd
@@ -37,7 +37,7 @@
                 :components ((:file "blockchain")))))
                                        
 (defsystem "emotiq/startup"
-  :depends-on (crypto-pairings)
+  :depends-on (crypto-pairings actors)
   :components ((:module source
                 :pathname "./"
                 :serial t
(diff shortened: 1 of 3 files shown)


src/emotiq.asd
```diff
@@ -37,7 +37,7 @@
                 :components ((:file "blockchain")))))
                                        
 (defsystem "emotiq/startup"
-  :depends-on (crypto-pairings)
+  :depends-on (crypto-pairings actors)
```
easye, line 40: @guitarvydas 

Adding this dependency isn't strictly related to removing watchdog timers, right?  

Rather this is necessary to build the system binary because of the use of `actors:install-actor-system`.   

Not going to hold up the commit for this reason, but would like you to comment so that my understanding is clear.

easye (changes requested):
Not requesting changes per se, but please respond to my question about the addition of the `actors` dependency to `emotiq/startup`.

src/emotiq.asd
guitarvydas: Yes, I think you are correct.  "Actors" was added when testing the build on gitlab (which originally failed on binary builds).  Not strictly for removing the watchdog.

PR #199, opened 2018-06-05
Dbm 0605b

dbmcclain:
Merge dev into dbm, change watchdog timer code to refer to user replaceable vector *WATCHDOG-HOOK*, which defaults to DEFAULT-WATCHDOG-FUNCTION.

This makes it easier to replace behavior as needed for business requirements.

### DIFF
--- a/src/Actors/actors-startup.lisp
+++ b/src/Actors/actors-startup.lisp
@@ -215,7 +215,8 @@ THE SOFTWARE.
   (declare (ignore ignored))
   (unless (directory-manager-p)
     (install-actor-directory)
-    (install-actor-printer)))
+    (install-actor-printer)
+    ))
 
 #-:lispworks
 (eval-when (:load-toplevel)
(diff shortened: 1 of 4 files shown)


guitarvydas (commented):
I've reviewed the changes.  The only significant changes were to actors.lisp regarding default-watchdog-function and the defmonitor function "check-sufficient-execs".  I see nothing wrong with these changes, although, personally I wouldn't bother with the special *watchdog-hook* and would expect developers to hack "default-watchdog-function" directly as business cases come and go, but, that goes to our discussion - I look for specialization and David looks for generalization.

There is also a change in line 527 adding :SBCL.

I approve, but, am punting to MHD and/or MarkE (added both to reviewer list) since they have been taking care of checking DBM into DEV.

Also, GitLab is failing on the binary builds, because "actors" is missing from depends-on in emotiq/startup.  Mark appears to know about this already.

markhdavid:
Looks fine to me. Will approve and rebase/merge.

markhdavid (approved):
approved

PR #200, opened 2018-06-05
added depends-on actors to emotiq/startup so that binary builds work

guitarvydas:
This commit simply add "actors" to the depends-on list of emotiq/startup.  Mark E already appears to know about this and might have a better solution.

GitLab has been failing everything today due to binary builds.  This commit should repair that.

I will close this PR if you have something different in mind...


markhdavid (approved):
yes

markhdavid:
approved, doing a rebase and merge

PR #201, opened 2018-06-06
Svs9

svspire:
Many, many changes to gossip.
Logging is now more ramified and ac:pr now goes to the `*log*` as well as to `*standard-output*`.
Added "dot printer" such that every time a log message is added to the log, a period is printed on `*standard-output*`. This will help debug hangs that happen in gossip tests in CI.
Slightly integrated with Cosi, in that both gossip and Cosi can be loaded at the same time and both still work. Cosi-BLS now depends on gossip.


echupriyanov:
> Added "dot printer" such that every time a log message is added to the log, a period is printed on standard-output. This will help debug hangs that happen in gossip tests in CI.

Can we also print log message to *standard-output* instead of dot? We don't have access to local files in CI. Though, we can upload them somewhere...

dbmcclain (approved):
Everything looks okay. Runs emotiq/sim just fine.

PR #202, opened 2018-06-06
Added `with-timeout.sh` to all test commands

echupriyanov:


easye (approved):
"ten minutes" should be enough for anyone…

PR #203, opened 2018-06-07
Svs10 mte

easye:


svspire:
pubkeys.conf is obsolete and should be deleted. Shouldn't be described in readme.

Now fixed.

svspire:
```
(defun start-node ()
   (gossip:gossip-startup :ping-others t))
```

should be

```
(defun start-node ()
   (gossip:gossip-startup :ping-others nil)
   ;(start-pinger-daemon) ; haven't written this yet but it will ping others periodically
; must work with David to figure out what to do with the information it returns.
  )
```

Now fixed.

svspire (approved):
Done. Didn't realize there was a config directory of its own and one *within* gossip. The readme within gossip/config is the latest. Other than the readme, you can blow both of these directories away.

PR #205, opened 2018-06-07
Fix portability of loenc strings

markhdavid:
- Unicode-string-p introduces nonportable differences in handling of
  strings of certain types. On LispWorks, if a string was not of type
  lw:8-bit-string, the string was assumed to have non-8-bit
  characters, even if it had none, whereas on other implementations it
  would check if there were some non-8-bit chars. The distinction
  results in different encoding, needlessly introducing implementation
  incompatibility.
- Solution was to use the Lispworks type test only as an optimization,
  but as definitive in the case of negative. Additional minor changes
  for performance: *char-marker* made a constant and test converted to
  highly efficient loop.


dbmcclain (approved):
looks interesting. I'll approve for now and see over time if we run into problems.

PR #207, opened 2018-06-07
Add string normalizer utility

markhdavid:
- This introduces base-string and simple-string as the normalized
  string types. When strings are created via Lisp READ, as well as
  when they're created in myriad ways through code, they can be of
  various types, such as simple-base-string, simple-string, or string,
  but these result in different encoding types, and this will result
  in different encoding using lisp object encoder system.  This change
  provides functions to coerce (copy if needed) and to simply copy
  unconditionally a string as an equivalent string of one of these two
  particular types, i.e., either BASE-STRING or SIMPLE-STRING.
- Note: the motivation for this was to deal with the practical issues
  arising from the fact that string literals with ASCII characters in
  SBCL are by default created as strings of type SIMPLE-STRING whereas
  in CCL and LispWorks they are of type BASE-STRING.  This results in
  lossage around loenc:encode/decode when applied to such strings
  going cross platform.

### DIFF
--- a/src/package.lisp
+++ b/src/package.lisp
@@ -7,7 +7,10 @@
   (:export #:octet #:octet-vector #:octet-vector-p
            #:make-octet-vector #:ovref
            #:octet-vector-to-hex-string #:hex-string-to-octet-vector
-           #:string-to-octet-vector)
+           #:string-to-octet-vector
+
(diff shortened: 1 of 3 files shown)


dbmcclain (approved):
interesting... you know much more about Lisp character encodings than I do. I know it can be a bowl of worms. Approved.

PR #208, opened 2018-06-07
Correction to string normalizer utility for simple-base-string case

markhdavid:
- to test: (asdf:test-system :emotiq)
  - runs lisp-unit test normalizing-strings-as-simple-strings
  - source file: src/tests/emotiq-test.lisp

----

- Both STRING and BASE-STRING can have fill pointers or be adjustable,
  not SIMPLE-STRING or any SIMPLE-BASE-STRING.

- Base-string can have fill pointers and be adustable. These would not
  be passable to/from C foreign functions, presumably, I don't
  think.

- So it was incorrect to have normalizers SIMPLE-STRING and
  BASE-STRING. Instead, it should have been SIMPLE-STRING and
  SIMPLE-BASE-STRING. So that's what this commit changes.

### DIFF
--- a/src/package.lisp
+++ b/src/package.lisp
@@ -9,7 +9,7 @@
            #:octet-vector-to-hex-string #:hex-string-to-octet-vector
            #:string-to-octet-vector
 
-           #:normalize-to-base-string #:copy-as-base-string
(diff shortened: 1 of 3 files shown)


dbmcclain (approved):
thanks for revisiting this question

PR #209, opened 2018-06-08
Start refactoring logging and configuration within gossip

easye:
* current `emotiq:main` behavior (sleep then ping)
* all messages go through`emotiq:note` facility for deciding what to do with logging
* logging starting to get refactored out of gossip (incomplete)
* configuration of gossip moved to `gossip/config`


src/config/config-files-readme.txt
```diff
@@ -1,40 +0,0 @@
-Config files:
```
guitarvydas, line 1: I see that you've removed a lot of stuff from .../config/... as you said you would.  I thought that we wanted the readme to be kept.  I see the readme being removed, but I don't see it being "kept" somewhere else?

guitarvydas (approved):

PR #211, opened 2018-06-10
The test harness may be used from an interactive shell again

easye:
To test, one should be able to invoke the test harness from an interactive shell via:

```
lisp=ccl ./etc/test-harness.bash
```

where the `lisp=ccl` may be replaced via setting a reference to the path of the Lisp which one wishes to run the test harness under.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,8 +4,8 @@ variables:
   LWPRO: lwpro
   LWPRO_MACOS_64: lwpro
   lisp: lisp-wrapper.sh
+# Add ASDF systems here to test them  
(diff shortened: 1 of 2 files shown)


easye:
@guitarvydas Please review this pull request now when you get a chance…

guitarvydas:
I put ccl on my PATH.  I ran ccl just to see that it works.  Then I tried `lisp=ccl ./etc/test-harness.lisp` and got `-bash: ./etc/test-harness.lisp: Permission denied` .  My files are owner=tarvydas, group=staff.  Groups says that I'm in staff.

guitarvydas (approved):
I can run test-harness.bash from a local command line.  I see that Pederson-proof tests are failing, but that is not the point of this PR.

PR #213, opened 2018-06-11
Fix pbc library references

easye:
The `CFFI:DEFINE-FOREIGN-LIBRARY` macro doesn't allow very sophisticated
forms within its expression.  Instead of specifiying an absolute path
within that macro, simply push the directory containing the libraries
into `CFFI:*FOREIGN-LIBRARY-DIRECTORIES*`.


easye:
Superseded by #215

PR #216, opened 2018-06-11
Split high-level GOSSIP and low-level GOSSIP/TRANSPORT 

lukego:
This branch spins the networking code from the GOSSIP module into a separate GOSSIP/TRANSPORT module.

GOSSIP/TRANSPORT is intended to be pluggable with multiple implementations. The initial implementation is based on threads and blocking operations (usocket.) (The next implementation will be non-blocking.)

TODO:

- [ ] Test and debug networked multi-node Gossip operation.


easye (commented):
It would be real nice to have `gossip/transport` ASDF definition that would allow the transport to be tested independently of the abstract Gossip implementation.  

I started to hack that but it isn't easy.  I was going for the removal of all references to 
```
(defvar *udp-proxy-table* (make-hash-table) "Hash table of all UDP proxies on this machine")
(defvar *tcp-proxy-table* (make-hash-table) "Hash table of all TCP proxies on this machine")
```
but that is not easy to do.

src/gossip/gossip-transport.lisp
```diff
@@ -0,0 +1,270 @@
...
+                                (let ((mailbox (mpcompat:make-mailbox)))
+                                  (setf (gethash (list address port) message-queues) mailbox)
+                                  (push (mpcompat:process-run-function (format nil "gossip/transport writer to ~A:~A" address port)
+                                                                       nil
+                                                                       (lambda ()
+                                                                         (run-tcp-write-thread transport address port mailbox)))
+                                        writers)
+                                  mailbox))))))
+
+(defun test ()
```
easye, line 242: TODO: move to separate source file unit under control of `asdf:test-system`

src/gossip/gossip.lisp
```diff
@@ -58,6 +58,8 @@
 
 (defvar *debug-level* 1 "True to log debugging information while handling gossip requests. Larger values log more messages.")
 
+(defvar *transport* nil "Network transport endpoint.")
```
easye, line 61: It would be better if we could somehow not have to touch `gossip.lisp` to boot the transport code, so that `gossip/transport` could be tested independently of `gossip`, but trying to hack this separation doesn't look trivial.

lukego:
@easye and @svspire: Do you want me to take any specific actions to revise this branch now?

easye (changes requested):
Please merge/rebase with `dev` in this branch, as there are a fair number of changes around `emotiq:main` startup as well as `gossip/config` execution that need to be accounted for to evaulate this code on its own.

PR #217, opened 2018-06-11
Dbm 0611a

dbmcclain:
Make Bulletproof basis vectors derive from successive hashes of the digits of Pi. This keeps everything publicly verifiable, yet safe, because no ECC point has a known discrete log.

Impact: None


easye (changes requested):
The change to `range-proofs:simple-commit` function signature causes the uncloaked transaction test in <https://github.com/emotiq/emotiq/blob/dbm-0611a/src/Cosi-BLS/t/cosi-tests.lisp#L66> to fail. 

You need–at a minimum–to modify `cosi/proofs:make-admin-commit` to adjust to the new four parameter version.

dbmcclain:
thanks for finding that. but sadly, once new transactions arrive, this code will all go away...

dbmcclain:
carried over Paul's fix to the missing age binding in default-handler, but without race conditions.

easye:
@dbmcclain You've added commits unrelated to the original pull request here.  

Could you please provide a description of what problem the commits <https://github.com/emotiq/emotiq/pull/217/commits/035876d04b8399db061f92a0bde59284a5c5f276> and <https://github.com/emotiq/emotiq/pull/217/commits/035876d04b8399db061f92a0bde59284a5c5f276> address, and how I may observe the change?

dbmcclain:
Paul tried to fix a missing lexical binding in the default handler for stalled Actors. His solution produced a race condition in that between the time the handler is spawned in its own thread and gets to run, another Executive thread might have managed to switch Actors, causing *last-heartbeat* to become updated. When handler finally runs, it could compute an age near zero and report that in the prompt.

Solution was to pass the already computed age from where the decision to spawn the handler was made, and make age a parameter of the handler call. Other handlers are free to ignore the parameter, or declare a &rest ignored.

easye (approved):
@dbmcclain Please try to keep pull requests grouped around the original thrust.  It causes real headaches for the reviewing process to have unrelated changes show up on the branch in the middle of reviewing.

PR #218, opened 2018-06-11
GraphID

svspire:
Add graphID so that gossip-nodes can be part of multiple graphs simultaneously.
Fix nasty CFFI bug by making calls to pbc:make-key-pair happen only once and inside an actor.
Lots of other smaller bug fixes and improvements.


guitarvydas (commented):
I'm getting "cannot create socket Error" in CCL.  An error I didn't get before, as far as I can remember.

easye (approved):
Verified that GOSSIP:PING-OTHER-HOSTS works on a two node local configuration.

PR #219, opened 2018-06-12
Initial support to run `devnet` Emotiq node in Docker

echupriyanov:


easye (approved):
`emotiq:main` runs fine.

PR #220, opened 2018-06-12
fix unbound "age" in default-watchdog-function

guitarvydas:
Should fix the problem of "age" being unbound in emotiq/src/Actors/actors.lisp/default-watchdog-function.

### DIFF
--- a/src/Actors/actors.lisp
+++ b/src/Actors/actors.lisp
@@ -634,6 +634,8 @@ THE SOFTWARE.
   #-(or :CLOZURE (AND :LISPWORKS :MACOSX)) 4)
 
 (defun default-watchdog-function ()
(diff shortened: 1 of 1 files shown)


dbmcclain:
Paul, I went ahead and fixed in my pull request. You had the right idea, but unfortunately, there becomes a race condition in which the handler might be called due to aging, yet some thread finally manages to switch Actors before the prompter thread gets to run the handler, causing the *last-heartbeat* to become updated, and your prompter message will indicate near zero seconds elapsed.

I fixed by passing age as a parameter, already computed in the decision to spawn the prompter. Other handlers are free to ignore the parameter, or just declare &rest ignored.

guitarvydas:
DBM's fix is better.

PR #221, opened 2018-06-12
Fix gossip/config external API

easye:
* properly export symbols used as the external interface
* eliminate compiler warnings


echupriyanov (approved):
Looks fine to me.
Only one suggestion to support several different ports on the same hostname

PR #222, opened 2018-06-12
157484804 create websocket interface

easye:


guitarvydas (changes requested):
I ql:quickloaded :emotiq/startup and got this error ...
```
UNSIGNED-BYTE which is referenced by EMOTIQ/WALLET::ENCRYPT-WALLET

(:EMOTIQ/STARTUP)
```
I'm not even sure that I ql'ed the correct .asd.

@echupriyanov @markhdavid - I think that I was included in the review of this PR because of logging.  I don't see anything wrong with the logging calls (but I can't run it).  I think you guys are much more familiar with REST and WEBSOCKETS, so I will punt to all y'all.  I can't approve this PR because it touches too many aspects that I have little experience in, but, I do put my stamp on the logging stuff, as long as the above error is addressed (maybe just needs an explanation, or a missing package name).

easye:
@guitarvydas I didn't actually request anyone to review this pull request, as it is not complete in its current form.  You, along with @markhdavid and @echupriyanov were "suggested" reviewers, not actually requested. 

I was using the pull request as a way to review my progress for myself alone.  Sorry for the confusion.

PR #225, opened 2018-06-14
157484804 create websocket interface

easye:


easye:
@guitarvydas and @echupriyanov 

This branch starts gossip, websockets, and REST.  The problem is that if a gossip configuration doesn't exist, the code will execute.  

It would be helpful if both of you could help me figure out what the right thing to do would be. 

If we can get this merged, Eugene can then query for wallet creation via a simple REST request to the running daemon.  

We need to figure out how to shutdown the node.

guitarvydas (commented):
I still see
```
[package hunchensocket]..
.........................
[package websocket/wallet].

UNSIGNED-BYTE which is referenced by EMOTIQ/WALLET::ENCRYPT-WALLET

(:EMOTIQ/STARTUP)
```
hmmm.  see http://clhs.lisp.se/Body/t_unsgn_.htm.  Lispworks.

easye:
@guitarvydas @markhdavid  Please review the 15-FRI-2018 updates.  

@echupriyanov This is the version of `emotiq` that you can communicate via REST/websockets to query/control the node for testing and deployment.

guitarvydas (approved):
I didn't have time to browse the websocket stuff.

PR #229, opened 2018-06-17
MTE fix simulator

easye:
If we define a working simulator as that which produces an observable block after `emotiq/sim:run` is executed, the simulator has been broken for cloaked transaction simulations since

<https://github.com/emotiq/emotiq/commit/2d2875ab1f7a05157b6f1efce8b700868bd7ff04>
```
git commit 2d2875ab1f7a05157b6f1efce8b700868bd7ff04
2018-05-29 08:46 -0700 David McClain <dbm@emotiq.ch>
Merge branch 'dev' into dbm

# Conflicts:
#	src/Cosi-BLS/cosi-netw-xlat.lisp
#	src/node-sim.lisp
```

The problem lies in the timeout of the prepare phase as controlled by ` cosi-simgen:*cosi-prepare-timeout*`.  This commit set the simulator to use a timeout of 10 seconds in waiting for the prepare phase with cloaked transactions.  During this time, all simulated nodes other than the leader must run the computationally expensive Bulletproof range proofs on amounts within the block.  If a Byzantine-fault tolerant threshhold of witnesses willing to commit are not received before this timeout occurs, the block is never sealed.  

This pull request changes the default timeout for the prepare phase to be 40 seconds which is enough for a 2013 8 core  Intel Core Duo 2 to simulate a four network node verifying cloaked transaction.  If need, the timout value may be set via the `:cosi-prepare-timeout` parameter to the `cosi/sim:initialize` function.

The simulator has been roundly violated since I last touched her, so I removed things that were no longer necessary, updated the docstring to reflect the true nature of the `emotiq/sim:run`, and introduced the `emotiq/sim:nodes` function to return a list of simulated nodes so that developers can inspect each nodes' version of the mempool, utxo, blockchain, etc. references.

### DIFF
--- a/src/election-sim.lisp
+++ b/src/election-sim.lisp
@@ -31,7 +31,8 @@ THE SOFTWARE.
 
 (defvar *beacon-interval*  20)
 
-;; ordered list/vector of all known stakeholders
+;; list of lists of a public key and its associated stake
+;; 
 ;; in real life, this should be an ordered list/vector of *real-nodes*
 ;; in the simulator, 1 machine, this is a list of all nodes (fake and real)
 ;; In the simulator, we only hold mock elections, and print the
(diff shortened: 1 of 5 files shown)


markhdavid (approved):
Looks good. In under 2 minutes of running uncloaked per the nicely improved standard doc, the 2 blocks existed and looked about right. Thanks for improvements to doc and package exports. Approved.

markhdavid:
Thanks for tracking down this weekend! Great detective work with git bisect!

PR #232, opened 2018-06-18
Pt new transactions2

guitarvydas:
Appears to remove 2 problems with new-transactions (1) duplicate genesis block (2) duplicated transactions.


src/Cosi-BLS/block.lisp
```diff
@@ -185,7 +185,7 @@ added to the blockchain."
             (cosi/proofs:public-key-to-address public-key)))
          (transactions (list genesis-transaction))
          (block (create-block nil nil nil nil transactions)))
-    (cosi-simgen:node-dispatcher :genesis-block :blk block)
+    ;(cosi-simgen:node-dispatcher :genesis-block :blk block)
```
markhdavid: I understand this is intended to eliminate duplicate genesis block problems, but now it does not correspond to the documentation. Don't we want the genesis block to be created on the genesis node, the originating node of the blockchain?  It seems that when you do that, the problem is that broadcast-message does not notice it's been done on that node already, and creates it anew.  Maybe that should be fixed. Or maybe the way it gets created is not done exactly right.  I think we should go back and fix it better.

src/Cosi-BLS/new-transactions.lisp
```diff
@@ -554,7 +554,8 @@ OBJECTS. Arg TYPE is implicitly quoted (not evaluated)."
                             (- sum-of-inputs sum-of-outputs)
                             *minimum-transaction-fee*)
                       (return nil)))
-                  (format t "~%Successful transaction ~a~%  TxID = ~a~%" 
+                  (format t "~%Node ~a declares Successful transaction ~a~%  TxID = ~a~%"
+                          (cosi-simgen::current-node)
```
markhdavid: The double-colon should not be used to reference cosi-simgen:current-node, which is an exported symbol.

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -1203,6 +1203,7 @@ check that each TXIN and TXOUT is mathematically sound."
     
 (defun leader-exec (prepare-timeout commit-timeout)
   (let ((trns  (get-transactions-for-new-block)))
+(format *standard-output* "~&leader sees transactions ~a~&" trns)
```
markhdavid, line 1206: Nonstandard formatting. However, I presume this is a transient debug/trace message.

src/Cosi-BLS/block.lisp
```diff
@@ -185,7 +185,6 @@ added to the blockchain."
             (cosi/proofs:public-key-to-address public-key)))
          (transactions (list genesis-transaction))
          (block (create-block nil nil nil nil transactions)))
-    ;(cosi-simgen:node-dispatcher :genesis-block :blk block)
```
markhdavid: (see comment in previous commit that commented this code out: I am not convinced this code should be eliminated, and now the doc is out of sync)

src/Cosi-BLS/new-transactions.lisp
```diff
@@ -1378,14 +1378,14 @@ ADDRESS here is taken to mean the same thing as the public key hash."
...
-        with transactions = (cosi/proofs:block-transactions block)
+        with transactions = (mapcar #'hash:hash/256 (cosi/proofs:block-transactions block))
         with block-tx-count = (length transactions)
         ;; initially (format t "~%Transactions count in block: ~d" (length transactions))
         ;;           (cosi/proofs/newtx:dump-txs :block block)
         for tx being each hash-value
           of mempool 
             using (hash-key key)
-        when (member tx transactions :test #'eq)
+        when (member (hash:hash/256 tx) transactions :test #'hash:hash=)
```
markhdavid: OK, so the logic of this change working is that the transactions in the block are not necessarily EQ to the transactions that are in the mempool.  I thought they were.  In get-transactions-for-new-block, the transactions are simply pulled out of the mempool to create the list of transactions, and that's what makes up the list of transactions in the block.  I need to figure out: how could they end up with a different identity?  Also, I understood in the code in question, it's only run in the case of the leader.

markhdavid (changes requested):
I put in some hopefully good feedback.  I definitely am not ready for this to go into dev.  The commits need to be squashed and prepared and cleaned up better (my changes as much as anyone's). But this is very useful and good work, and thanks to Paul for much hard work on this. David also worked on this a bunch today, and Mark E a bunch over the week. Thank you all again.

src/Cosi-BLS/new-transactions.lisp
easye: @markhdavid In every node other than the leader, the block containing the transactions has been de-serialized from the network.  Therefore `eq`, which tests object identity for the references to the same block of memory, isn't going to be equal.

src/Cosi-BLS/block.lisp
easye: The genesis block is not necessarily created by a single node.  It is simply a valid block that all nodes can begin their computation from.  To actually have a leader orchestrating the preparation, witnessing, and signing the block via network communication is probably more complicated than it needs to be.

PR #235, opened 2018-06-19
Gossip API

svspire:
Gossip API for integration with COSI is mostly done. Can't establish ad-hoc groups yet, but we can use the special :UBER graph which just uses all known nodes/keys, and this is enough for MVP.
Some improvement of startup/shutdown of gossip via graceful-shutdown and gossip-final-cleanup.


easye (commented):
Simple two node connectivity test fails:

```
#(254 128 0 0 0 0 0 0 104 91 53 255 254 219 186
  100) fell through ETYPECASE expression inside USOCKET::HOST-TO-HBO.
Wanted one of (STRING (OR (VECTOR T 4)
                          (ARRAY (UNSIGNED-BYTE 8) (4)))
                      INTEGER).
   [Condition of type CONDITIONS:CASE-FAILURE]

```

Looks like you are getting bytes representing IP6 addresses when you convert from names under some conditions (i.e. when symbolic names are configured as part of `gossip/config`).  You should probably wrap conversion to IP4 in an appropriate handler.

easye (approved):
Both `emotiq:main` without configuration and the simulator continue to run.  

Approved.

guitarvydas (approved):
works for :emotiq/sim
for :emotiq/startup (emotiq:main)
(ql:quickload :gossip), (gossip:gossip-startup), (gossip:ping-other-machines).

PR #236, opened 2018-06-19
Introduce new transactions (smart contract, with 'chainlisp').

markhdavid:
New transactions are on by default in this commit, but old code continues to
live side by side for now and is switched on or off via switch parameter
cosi/proofs:*newtx-p*. The parameter is true by default, meaning new
transactions are enabled by default.

Module node-sim.lisp has been updated to have new run function run-new-tx which
can only work with new transactions. It uses new functionality to create a
genesis block (no longer a "genesis UTXO").  It includes a few very basic spend
transactions as well as a couple of intentional attempted double-spend
transactions that should be rejected.

As the current test stands (EMOTIQ/SIM:RUN-NEW-TX) runs and thinks for about a
minute, then calls a new transaction "dumper" to display all transactions on the
blockchain so far, showing there are two blocks with a total of four
transactions.  Other output shows two rejected attempted double-spends.

This commit introduces a new module new-transactions.lisp and a new package
COSI/PROOFS/NEWTX, intended to be temporary, in which the class transaction is
defined (the old transactions are defined with the same name in the package
COSI/PROOFS).  Later, work will be done on decommissioning the old transactions,
and presumably on rethinking names, e.g., not to contain "new", etc.

Logging/messaging/signaling are at this point a hodge-podge. This commit does
introduce the emerging emotiq:note facility, and begins using it to
log/display/warn about things, but there numerous are other conventions or lack
thereof at work.

Expect cleanups, forthcoming. Highly notably missing from this commit and
needed/planned for subsequent commits: cloaked transactions; other types of
transactions: stake, etc.; some of the types of validation that was
lost/commented out in this commit (with comments); creation of coinbase
transaction at beginning of each block and computing and including fees,
including fees; using priority age|fees to prune transactions from mempool for
block if too many; including input scripts (including witness data) in block
(i.e., block header) for ultimate validation; UTXO database; blockchain on disk
(persistent storage); genesis block creation is still wild west; etc. (This list
is not necessarily exhaustive.)

### DIFF
--- a/src/Cosi-BLS/block.lisp
+++ b/src/Cosi-BLS/block.lisp
@@ -165,6 +165,31 @@ added to the blockchain."
       blk)))
 
 
+
+(defparameter *newtx-p* t) ; using new-transactions.lisp (package
+                           ; cosi/proofs/newtx) - MHD development!
+                           ; -mhd, 6/13/18
+
+
+
+(defun create-genesis-block (public-key)
+  "Create a genesis block on cosi-simgen:*BLOCKCHAIN* paying to PUBLIC-KEY. This
+   must be called on the genesis node of the blockchain, with
+   cosi-simgen:*CURRENT-NODE* appropriately bound. Note that the funds are sent
+   to a hash of the pubiic key, converted with
+   cosi/proofs:public-key-to-address."
+  (when (not *newtx-p*)
+    (error "Sorry, only know how to do this for new transactions (*newtx-p*)."))
+  (let* ((genesis-transaction
+           (cosi/proofs/newtx:make-genesis-transaction
+            (cosi/proofs:public-key-to-address public-key)))
+         (transactions (list genesis-transaction))
+         (block (create-block nil nil nil nil transactions)))
+    block))
+    
+             
+
+
 (defun hash-256 (&rest hashables)
   "This is the hashing used in our block in a merkle tree, linking
    transaction outputs and inputs, and hashing the block header. This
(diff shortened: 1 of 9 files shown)


easye (commented):
@markhdavid A little too squashed for my tastes, as I would like to see discrete developer's contributions as part of the commits (e.g. that <https://github.com/emotiq/emotiq/pull/234/commits/8bc14bcabf538b2660e747e049584418ff5e3366> is present) without having to hunt through the total squashed diff.  But I ain't gonna refuse this request on that basis, just noting that such a squash causes me to redo the previous work I did today in checking <https://github.com/emotiq/emotiq/pull/234>

easye:
A little confusing to have `emotiq/sim:run` *and* `emotiq/sim:run-new-tx`.  

Recommend having only one exported symbol, or else we need to maintain both simulations of the old transaction and the new one.   Probably should just use `emotiq/sim:run` as this point.

easye:
@guitarvydas In the near future, we're going to need the simulation to run as part of `emotiq:main` as well as booting up `gossip` in order to run automated tests on the nodes.

markhdavid:
The bug fix is not very useful relative to dev. On dev we should have an initial commit that introduces the feature.  On dev prior to this commit, there never was this, but so there is no use introducing the bug and then having the individual fix for it.

easye (approved):
Behaves like <https://github.com/emotiq/emotiq/pull/234>.  

Approved.

easye:
>The bug fix is not very useful relative to dev. On dev we should have an initial commit that introduces the feature. On dev prior to this commit, there never was this, but so there is no use introducing the bug and then having the individual fix for it.

The issue is not that the bug fix is not useful relative to dev, but that I had to manually hunt through the diff to verify it was present.

markhdavid:
Mark E, I know, but the stuff with "new" is a necessary evil of having a transitional period with things with "new" in their name. So, it's fair to have the criticism you're making, but we have to live with it for a while.

src/Cosi-BLS/block.lisp
```diff
@@ -165,6 +165,31 @@ added to the blockchain."
       blk)))
 
 
+
+(defparameter *newtx-p* t) ; using new-transactions.lisp (package
```
easye, line 169: @guitarvydas Please ensure that we deprecate `*newtx-p*` as soon as conveniently possible, possibly by confirming when @dbmcclain is ready for it to be removed.  Probably, simultaneously coordinate the replacement of the `emotiq/sim:run` routine with `emotiq/sim:run-new-tx`.  

Such conditional paths in the source code based on a special at the point when we have no need of "old transactions" are a legendary basis for runtime errors.

PR #241, opened 2018-06-20
Clean up initial commit of new transactions around mempool clearing.

markhdavid:
Added remove-transaction-from-mempool, which knows to use
get-hash-key-of-transaction to get its key. That eliminates whole
reason for the bug earlier (while preparing the previous commit) of
using the old-transactions-style key (completely my fault), which
required the painful and clever coding workaround that PT figured out,
essentially rehashing both the transactions in the block and the
transactions in the mempool and then comparing their hashes using
hash=.

Hopefully, it's clearer now with this change.  Added an error message
by node-check-transaction in the case of *newtx-p* just for emphasis.

To test: run with newtx mode, see the right number of Tx's (4)
dumped. To test old: restart, and run with old tx mode, see right
number Tx's (2) eventually built.

### DIFF
--- a/src/Cosi-BLS/cosi-handlers.lisp
+++ b/src/Cosi-BLS/cosi-handlers.lisp
@@ -388,6 +388,8 @@ Later it may become an ADS structure"
   nil)
 
 (defmethod node-check-transaction ((msg transaction))
+  (when *newtx-p*                       ; sanity check
+    (error "This method must not be called in new-transactions mode."))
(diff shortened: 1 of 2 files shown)


markhdavid:
Have asked (on Twist) for @guitarvydas to primarily take the lead on reviewing, and he has said will do so presently.

guitarvydas (approved):

PR #244, opened 2018-06-20
fire only one election - remove election beacon, call fire-election a…

guitarvydas:
…fter all txns have been broadcast ; result = 2 blocks (genesis, plus block with 3 txns)

This should run the simulator and execute exactly one election.  The Leader is determined at runtime.

The result should be 2 blocks.  The first block is the genesis block.  The 2nd block contains the 3 valid transactions.  Run-new-tx creates 5 transactions (plus genesis), but 2 of them are deliberate double-spends and are discarded

commands: (ql:quickload :emotiq/sim) (emotiq/sim:initialize) (emotiq/sim:run-new-tx)

This 'one-shot' election should be useful when debugging and when bringing the network up the first time.

@MarkD - sorry, this is probably what you wanted earlier, but it didn't occur to me until David showed me the trick.

### DIFF
--- a/src/node-sim.lisp
+++ b/src/node-sim.lisp
@@ -205,7 +205,6 @@ This will spawn an actor which will asynchronously do the following:
   (ensure-simulation-keys)
   (setf *genesis-output* nil *tx-1* nil *tx-2* nil)
   (cosi-simgen:reset-nodes)
-  (emotiq/elections:make-election-beacon)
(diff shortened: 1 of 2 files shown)


markhdavid (approved):
Thanks for adding this, since it's going to be helpful for debugging.

easye (approved):

PR #245, opened 2018-06-21
Notes are not errors.

svspire:
Add `*notestream*` as output of emotiq:note function. Initialize it to
`*error-output*`, but allow it to be rebound dynamically (even to nil
for no output at all) without being required to change the connotation
of `*error-output*` as "the stream for errors."

Get rid of `*log-dots*` and `*logstream*` since we're no longer using them.

### DIFF
--- a/src/gossip/gossip-startup.lisp
+++ b/src/gossip/gossip-startup.lisp
@@ -104,10 +104,14 @@ with this list if desired."
        #+IGNORE ; #+OPENMCL ; these hemlock streams are just too slow when they get to a couple thousand lines
        ; in CCL, we'll just inspect the *log*
        (if (find-package :gui)
-           (setf *logstream* (funcall (intern "MAKE-LOG-WINDOW" :gui) "Emotiq Log"))
-           (setf *logstream* *standard-output*))
-       #+(or (not :openmcl) (not :shannon))
-       (setf *logstream* *error-output*)
+           (setf emotiq:*notestream* (funcall (intern "MAKE-LOG-WINDOW" :gui) "Emotiq Log"))
+           (setf emotiq:*notestream* *error-output*))
+       (setf emotiq:*notestream* 
+             #+(or (not :openmcl) (not :shannon))
+             *error-output*
+             #-(or (not :openmcl) (not :shannon))
+             nil
+             )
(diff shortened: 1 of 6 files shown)


easye (changes requested):
After `gossip-tests` end to run under CCL, one gets errors 

```
Invalid SEND target: 
  NIL
   [Condition of type ACTORS:INVALID-SEND-TARGET]

Restarts:
 0: [TERMINATE-ACTOR] Terminate Actor
 1: [ABORT-BREAK] Reset this thread
 2: [ABORT] Kill this thread
```

and setting your heartbeat timers to the default causes watchdogs to signal stalled conditions as well

```
Actor Executives are stalled (blocked waiting or compute bound). 
Last heartbeat was 3 sec ago.
   [Condition of type SIMPLE-ERROR]
```

This is a regression that is also present on `dev`, but is related to the logging facility.

Please ensure that executing `gossip-tests` does not signal conditions.

guitarvydas:
@MarkE (@Shannon) to the best of my recollection, the "Actors Stalled" problem went away when EC made some changes to the test setup (more cores?).  FYI - I encountered "Invalid SEND target" in David's code, back when he was registering and unregistering reply-to actors.  The problem in David's code went away, but maybe something similar is now happening in another part of the system.

echupriyanov:
@guitarvydas I've reverted number of cores back to 2, so issue went away for some other reasons. Also, I've added some defensive measures placing timeout on shell command execution, so dropping to debug REPL no longer blocks CI runner forever.

shannon:
@guitarvydas I think you @'d the wrong Shannon. I'm not familiar with this project.

guitarvydas:
^ @svspire @easye 
(yoiks, I've been @'ing other people, I intended to leave comments (above) for you guys)

svspire:
"Invalid SEND target" was caused by calling `(gossip::gossip-init ':uninit)` in `teardown-fixture`. This caused the `*logging-actor*` to be nullified when there were still log messages to be printed, and it killed the `*hmac-keypair-actor*` as well. I've modified `teardown-fixture` to do nothing now. Working in lg-svs3 branch.

svspire:
Closed. Now incorporated in lg-svs3-pr.

PR #246, opened 2018-06-21
Lg svs3 pr

svspire:
More bugs fixed in new gossip-transport integration.


src/gossip/gossip-startup.lisp
```diff
@@ -104,10 +104,14 @@ with this list if desired."
        #+IGNORE ; #+OPENMCL ; these hemlock streams are just too slow when they get to a couple thousand lines
        ; in CCL, we'll just inspect the *log*
        (if (find-package :gui)
-           (setf *logstream* (funcall (intern "MAKE-LOG-WINDOW" :gui) "Emotiq Log"))
-           (setf *logstream* *standard-output*))
-       #+(or (not :openmcl) (not :shannon))
-       (setf *logstream* *error-output*)
+           (setf emotiq:*notestream* (funcall (intern "MAKE-LOG-WINDOW" :gui) "Emotiq Log"))
```
guitarvydas: I like the idea of notestream and getting rid of dots and logstream, but ...

src/gossip/gossip-startup.lisp
```diff
@@ -104,10 +104,14 @@ with this list if desired."
...
+           (setf emotiq:*notestream* *error-output*))
+       (setf emotiq:*notestream*
```
guitarvydas: To my eyeballs, if +lispworks, we get a setf with 3 args.  I haven't tried it.

guitarvydas:
In my ATE stuff, I think I need a graceful-actor-shutdown call, too.  I would like to be able to shutdown the actors system without just killing all of the Executives (the Executives should commit suicide when they see no more work to be done).  I've opened this up to David (since he might have some interesting suggestions) and to MarkE (since he works the "shift" after me).  Your comments seem to indicate that you've found a work-around, but I'd be happier with a graceful shutdown.

easye (changes requested):
@svspire Please address my concerns about your use of conditionals as written up in <https://github.com/emotiq/emotiq/pull/245>

src/gossip/gossip-startup.lisp
```diff
@@ -37,7 +37,7 @@
               (make-node :uid pubkey))
             pubkeys)
       ;; clear log and start server
-      (run-gossip-sim :TCP)
+      (run-gossip-sim)
```
easye: @svspire The name `run-gossip-sim` is no longer accurate as it no longer refers to a simulation but rather the main loop.

easye (changes requested):
After the `gossip-tests` are run, there shouldn't be a server listening on `tcp:65002`.  

Please fix the `teardown-fixture` in the tests so that the tcp server is cleanly shutdown, and all gossip specific actor threads have been ended.

src/gossip/gossip.lisp
```diff
@@ -70,6 +70,31 @@
...
+;; pbc:signed-message, check for valid signature, then call user's
+;; handler with embedded authenticated message. If any failure along
+;; the way, just drop the message on the floor.
+
+(defun do-process-authenticated-packet (deserialize-fn body-fn)
+  "Handle decoding and authentication. If fails in either case just do nothing."
+  (let ((decoded (ignore-errors
+                   ;; might not be a valid serialization
+                   (funcall deserialize-fn))))
+    (when (and decoded
```
easye: Dropping messages without logging and calling `ignore-errors` without handlers installed produces code that will be very brittle and fail easily when it encounters messages that it does not process correctly for some reason.

easye (changes requested):

svspire:
>  Please address my concerns about your use of conditionals as written up in #245

I couldn't find any concerns about my use of conditionals.

src/gossip/gossip-startup.lisp
svspire: Fixed.

src/gossip/gossip-startup.lisp
svspire: Fixed.

src/gossip/gossip.lisp
svspire: Fixed.

svspire:
(ac::kill-executives) will kill actor threads. It's probably safer than killing them manually since it unschedules the heartbeat-timer as well.

svspire:
This PR should be ready for acceptance now. I'm still working on a race condition in #'hmac-keypair but that won't prevent the tests from running.

easye (commented):
@svspire Please write a docstring as to what how you intend to configure setting for the `emotiq/logging:*notestream` as the intention is unclear from the mess of `setf` and conditionals in the submitted code.  

All conditionals should be written so that an implementation other than we currently test could plausibly have a default implementation or value to attempt compilation. 

If neither `:openmcl` nor `:lispworks` were present in `cl:*features*`, what should this code do?

src/gossip/gossip-startup.lisp
```diff
@@ -96,31 +96,33 @@ with this list if desired."
...
     (case cmd
       (:init
        #+IGNORE ; #+OPENMCL ; these hemlock streams are just too slow when they get to a couple thousand lines
        ; in CCL, we'll just inspect the *log*
        (if (find-package :gui)
-           (setf *logstream* (funcall (intern "MAKE-LOG-WINDOW" :gui) "Emotiq Log"))
-           (setf *logstream* *standard-output*))
+           (setf emotiq:*notestream* (funcall (intern "MAKE-LOG-WINDOW" :gui) "Emotiq Log"))
+           (setf emotiq:*notestream* *error-output*))
        #+(or (not :openmcl) (not :shannon))
```
easye: Please write a comment as to what you intend to do with the logging ouput from `emotiq/logging`, as the intentions are unclear from the code as written, and I don't think this does what you want.

As I understand this conditionals, if the feature :shannon is not present you unconditionally muffle emotiq:note for all users if not running on CCL. If the feature :shannon is present on a Lisp other than CCL, you set effectively have

(setf emotiq:notestream error-output nil)

which maybe fails to compile, and certainly might behave non-standardly.

Part of clarifying things would be to remove :shannon to reflect semantics other than the current developer is not Shannon. Doesn't it mean :running-ccl-hemlock or :using-hemlock?

easye (changes requested):

src/gossip/gossip-startup.lisp
svspire: Mark, I didn't add the conditional for :SHANNON; you did. So I'm confused as to why you're confused.
https://github.com/emotiq/emotiq/commit/30bc299aa2888b0f79feb9f04b1effcf6e466dcf

Furthermore, the line `(setf emotiq:notestream error-output nil)` does not exist, and the code compiles just fine.

There are two points I'm trying to address herein:
1. Specifying `*error-output*` as the output stream for note messages is the wrong thing. Notes are not errors. Note output needs to go to its own stream whose connotation is "note" and which can be rebound as needed.
2. In CCL using the IDE on a Mac, every output written to `*standard-output*` or `*error-output*` gets sent to a listener associated with the current thread. In multithreaded code, this results in a complete mess on the screen with a separate listener window popping up whenever a new thread prints a note message. Therefore, in CCL using the IDE on a Mac, it's *very* important that the output of `note` not go to `t` or `*standard-output*` or `*error-output*`. In my case, I prefer it to be bound to nil, so I don't see any note output whatsoever. I have my own logging system in gossip which sends all messages to `gossip:*log*`, which is a vector of lisp objects whose events have accurate timestamps and are guaranteed to be properly serialized.

easye: I have not said that the code does not compile.

I have asked you to document the meaning of the features in the code

easye: @svspire 

`*error-output*` is exactly the right stream to use.  From CLHS

```
The value of *error-output*, called error output, is a stream to which warnings 
and non-interactive error messages should be sent.
```

I would then rename the `:shannon` feature to `:ccl-with-ide`, and ensure that conditionals work in the case when CCL is not being used with the IDE which should be the default without configuration as it is the case for other developers using SLIME, and for the CI system.

svspire: All fixed now.

easye (approved):

PR #250, opened 2018-06-25
Remove actor-based atomicity from hmac-keypair since it's designed wrong

svspire:
(see https://github.com/emotiq/research/wiki/A-Danger-with-Actors) and
it's no longer necessary since we're initializing PBC correctly now.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,7 +5,7 @@ variables:
   LWPRO_MACOS_64: lwpro
   lisp: lisp-wrapper.sh
 # Add ASDF systems here to test them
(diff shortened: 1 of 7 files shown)


guitarvydas (changes requested):
On 2 out of several runs, I got an assertion failure in sign-message.  I don't yet know if this is my environment or a real bug.  I'm running (ql:quickload :gossip), (gossip:gossip-startup), (gossip:ping-other-machines).  I'm running this with LWM.  I have the broken run on my screen and can send you more details.  (I am running on my own set of aws instances - I have done nothing to update them (?)).

guitarvydas (approved):

svspire:
Now just using a lock.

PR #252, opened 2018-06-25
Finishing spend transactions - verify transactions on nodes

markhdavid:
- verify transactions on nodes (check merkle root against recomputed
   merkle tree on each non-leader node). Now check-block-transactions
   does the checking and has documentation as to its purpose.

 - in addition, this commit adds the beginnings of generality for
   other types of transactions besides uncloaked spend transactions:
   first other type we'll use will be collect, and the next on after
   that will be cloaked spend. (Note that these are just no-op
   transformations that add some functionality [change that to FLEXIBILITY. -mhd] but should not end up
   changing any functionality as of yet.)

 - debugging improvement: added wait-for-tx-count and use in node-sim:
   this lets us sleep for the minimal amount of time waiting for the 4
   transactions to complete while showing "ticks" every second, an
   improvement over simply going to sleep for 60 seconds
   unconditionally.

### DIFF
--- a/src/Cosi-BLS/new-transactions.lisp
+++ b/src/Cosi-BLS/new-transactions.lisp
@@ -23,6 +23,37 @@ OBJECTS. Arg TYPE is implicitly quoted (not evaluated)."
 
 
 
+;;;; Transaction Types
+
+;;; A `transaction' in Emotiq Lisp software is implemented an instance of class
+;;; TRANSACTION. The Lisp type of an instance of this class is TRANSACTION. In
+;;; the following discussion, however, the word "type" is used in its generic
+;;; meaning, not Lisp/programming-language sense.
+
+;;; Emotiq has six types of transactions:
+;;;
+;;;   - Spend transaction to transfer tokens, with two subtypes
+;;;     - cloaked
+;;;     - uncloaked
+;;;   - Collect transaction for collecting block fees
+;;;   - Stake transaction for sending tokens into PoS escrow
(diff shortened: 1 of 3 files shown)


markhdavid:
If anyone approves this, please just approve, but then do not do a commit.  I want to revise the commit log message slightly.

markhdavid:
Aha, Paul did show me that (unfortunately) "all checks have passed" can be a false positive. And actually there are still failing tests, which he's told Mark E and me about in various messages elsewhere. Accordingly, I'll retract my approval again.

dbmcclain (approved):
I don't see any uses of gamma in this code. Just one mention of gamma in one comment.
I find the sieve of code presentation in GitHub to be too sparse to make any sense of what is shown, especially since none of the code seems to be rewriting anything with which I have intimate familiarity. So, on the presumption that you have tested the code, what more can I offer? Provisional approval granted...

markhdavid:
@dbmcclain yeah sorry this really did not relate to the "gamma topic" in that discussion

markhdavid:
Due to Github's incompetence amending commit messages in the midst of a PR (see https://github.com/isaacs/github/issues/998 and numerous googlable discussions), I created a new branch (154675977-smart-contract-transactions-21june2018-rebased-squashed-revise-commit-message) and new PR #253 simply to amend the commit log to replace one word with another (as noted in the top message above, see message in brackets []), and the commit was moved to dev from that PR on the authority of this PR's approval.

PR #253, opened 2018-06-25
Finishing spend transactions - verify transactions on nodes

markhdavid:
- verify transactions on nodes (check merkle root against recomputed
   merkle tree on each non-leader node). Now check-block-transactions
   does the checking and has documentation as to its purpose.

 - in addition, this commit adds the beginnings of generality for
   other types of transactions besides uncloaked spend transactions:
   first other type we'll use will be collect, and the next on after
   that will be cloaked spend. (Note that these are just no-op
   transformations that add some flexibility but should not end up
   changing any functionality as of yet.)

 - debugging improvement: added wait-for-tx-count and use in node-sim:
   this lets us sleep for the minimal amount of time waiting for the 4
   transactions to complete while showing "ticks" every second, an
   improvement over simply going to sleep for 60 seconds
   unconditionally.

### DIFF
--- a/src/Cosi-BLS/new-transactions.lisp
+++ b/src/Cosi-BLS/new-transactions.lisp
@@ -23,6 +23,37 @@ OBJECTS. Arg TYPE is implicitly quoted (not evaluated)."
 
 
 
+;;;; Transaction Types
+
+;;; A `transaction' in Emotiq Lisp software is implemented an instance of class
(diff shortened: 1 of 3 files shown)


markhdavid:
Same as #252 (which was approved) except one word wanted to replaced in the commit message, but since github notoriously cannot handle a force push in the middle of a PR, a new branch had to be created and a new PR created which is this (#253), which I'll consider approved.

PR #254, opened 2018-06-26
Fix major bug in new transport layer 

svspire:
where local gossip port wasn't getting sent or received in messages.
Fix minor bug in API.


echupriyanov (approved):
Looks fine for me

PR #255, opened 2018-06-26
watching CI to see if this fixes the multiple defvar problem

guitarvydas:
no PR - just checking gitlab on this


guitarvydas:
Nope - removing a multiple defvar for *default-wallet-name* (once in wallet.lisp and once in name.lisp) hasn't cured the problem.  The CI shows a green checkmark, but scrolling backwards in test:macos:lispworks shows an error (dev build)

PR #256, opened 2018-06-26
Dbm initial adoption of gossip

dbmcclain:
Initial adoption of true Gossip protocol. Changes have been minor and there is a switch var *USE-REAL-GOSSIP* to switch between old sim code and new Gossip based code. 

Accessor APPLICATION-HANDLER added to Cosi Node Self slot to help Gossip layer route incoming messages to app layer.

My local GOSSIP-SEND now calls GOSSIP:SINGLECAST, and my local GOSSIP-NEIGHBORCAST now calls on GOSSIP:BROADCAST.

(there is also some older changes to Edwards curves - just cleaning out old comments.)


dbmcclain:
Interesting question... Gossip messages are being BROADCAST to all witnesses in the Cosi group. (For now that is short-circuited to :UBER graph for initial tests, as per Shannon advice). But incoming blocks being validated contain a list of witness keys. 

So it seems we ought to be constructing Gossip graphs specific to the block's notion of witnesses instead of all available nodes in the Gossip overall graph. 

But, as I understand things, tearing down and setting up Gossip graphs involves some non-trivial network negotiations. You probably want to avoid making changes to this too often.

At any rate, code for adapting to Block's notion of witness pool still needs to be added. But for now, this code should run on the provisional Gossip network.

easye (changes requested):
This branch seems to fail to compile:

<https://gitlab.aws.emotiq.ch/emotiq/emotiq/-/jobs/3758>
```
[package cosi-test]...........................Read error between positions 33060 and 33307 in /builds/emotiq/emotiq/src/Cosi-BLS/cosi-handlers.lisp.
Failed to Quickload dependencies for system `COSI-BLS-TESTS`:
Reader error: No external symbol named "ESTABLISH-GOSSIP-BROADCAST-GROUP" in package #<Package "GOSSIP"> .
```

dbmcclain:
Apparently the Gossip API isn't fully ready for prime time. For now, just commented out section of code calling undefined external API symbol.

dbmcclain:
Uncommented after switching over to Shannon's GOSSIP:ESTABLISH-BROADCAST-GROUP. Apparently, Shannon doesn't want us to be too specific, preferring instead, to let his code determine best kind of group. 

This code isn't being called right now anyway, with the short-circuit of :UBER group, during testing. But at least the code loads and compiles cleanly for when we take out the circuit breaker.

dbmcclain:
Suggestion... when we produce Lisp source files labeled ...-api.lisp, to enclose some form of public API, either comment out sections that are meant to be private to the code and not part of the API, or else move private functions to another source file. 

As is, the gossip-api.lisp does not clearly state that establish-gossip-broadcast-group should not be called directly. Only find out when attempting to load and the symbol has not been exported from gossip package.

Please don't leave open manholes in the street for people to step into...

dbmcclain:
Maybe we even need an API macro that indicates which functions should be used by public clients and automatically does the symbol export?

easye (changes requested):
Approved.  

@markhdavid To take care of CCL compiler problem with `new-transactions.lisp`

easye (approved):

PR #257, opened 2018-06-26
158634075 generate gossip config port

easye:
Distinguish generated testnet configuration by port number.

### DIFF
--- a/src/gossip/config/generate.lisp
+++ b/src/gossip/config/generate.lisp
@@ -15,7 +15,9 @@ The ROOT defaults to '/var/tmp/conf'."
       (destructuring-bind (host eripa port (public private))
           key-record
         (declare (ignore private))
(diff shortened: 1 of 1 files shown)


echupriyanov (approved):
Thanks!

svspire:
Colons are illegal characters in filenames in MacOS. Is this only for Linux?

echupriyanov:
Why? They work ok in macOS

svspire:
Got to the Finder, click on a file, hit RETURN, and attempt to rename a file by adding a ":" character. It will fail.

svspire:
Historically -- prior to OSX -- MacOS used the colon as a directory separator. They have kept that ability for backwards compatibility. If you force a colon character into a filename programmatically, it will be internally escaped, which means the filename may or may not be portable to other file systems.

PR #258, opened 2018-06-26
add 'from 0' to loop for/as clause (required, LispWorks didn't enforce)

markhdavid:
fixes regression in CCL, including CI tests


markhdavid:
CCL would error on this code whereas LispWorks would handle it:
```
(loop for foo when (> foo 3) return foo)
```
LispWorks handles it just like this code:
```
(loop for foo from 0 when (> foo 3) return foo)
```
LispWorks is lenient. I'm not sure that's wrong.  CCL is strict. I'm not sure that's wrong. My code had been wrong, I'm pretty sure. *Sorry!*

easye (approved):
Verified compilation under CCL, thanks for the prompt fix.

PR #260, opened 2018-06-27
changes for fees and collect transaction, with other cleanups, fixes

markhdavid:
Working collect transactions, as well as cleanup of several other things.

Code to support making new collect type transactions, including code
to compute the fee on a transaction.

Made up a new type for the initial coinbase transaction in the genesis
block: :coinbase

Improved code to dump transactions, leaving out the useless
pseudo/dummy inputs in the case of collect and coinbase
transactions, and noting the transaction type in the case of
unusual, i.e., non-spend type, transaction.

Wait-for-tx-count, used by demo/test code in node-sim, had
various changes in response to the new types of transactions. Now
it counts only spend transactions by default. There's an option
to override/specify which types of transaction are counted.

The node-sim file's run-new-tx demo now advises that it's waiting
for 3 spend transactions, and then that it has finished 3 spend
transactions.  At the end, the dumper shows all the types of
transactions.  The dumper now identifies specially any
transaction that is not of type :spend, which is the most common,
standard, and default type.

### DIFF
--- a/src/Cosi-BLS/new-transactions.lisp
+++ b/src/Cosi-BLS/new-transactions.lisp
@@ -50,6 +50,7 @@ OBJECTS. Arg TYPE is implicitly quoted (not evaluated)."
 ;;;   :spend -- for an uncloaked spend transaction
 ;;;   :spend-cloaked -- for a cloaked spend transaction
 ;;;   :collect -- for a collect transaction
+;;;   :coinbase -- for the (one) transaction that creates coin (a/k/a the genesis transaction)
 
 
 
@@ -585,7 +586,7 @@ OBJECTS. Arg TYPE is implicitly quoted (not evaluated)."
                             (- sum-of-inputs sum-of-outputs)
                             *minimum-transaction-fee*)
                       (return nil))
-                    (emotiq:note "~%Node ~a declares Successful transaction ~a~%  TxID = ~a~%"
+                    (emotiq:note "~%Node ~a declares successful transaction ~a~%  TxID = ~a~%"
                                  (cosi-simgen:current-node)
                                  transaction
                                  (txid-string (transaction-id transaction)))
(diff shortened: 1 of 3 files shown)


src/Cosi-BLS/new-transactions.lisp
```diff
@@ -1389,35 +1380,42 @@ ADDRESS here is taken to mean the same thing as the public key hash."
 
 (defparameter *wait-for-tx-count-tick-print* ".")
 (defparameter *wait-for-tx-count-tick-every* 10)
-(defparameter *wait-for-tx-count-message* "Waiting for tx count: ")
+(defparameter *wait-for-tx-default-tx-types* '(:spend :spend-cloaked))
 
-(defun wait-for-tx-count (n &key interval timeout node
-                                 tick-print tick-every message)
+(defun spend-tx-types-p (tx-types)
+  (equal '#.(sort (copy-list '(:spend :spend-cloaked)) #'string<)
```
dbmcclain, line 1386: Sure this shouldn't be a membership test instead of a list equality test?

dbmcclain (commented):
Looked at use case for SPEND-TX-TYPES-P and I think that should be a membership test. Right now it appears to be a list equality test, meaning that you must submit a list of exactly :SPEND and :SPEND-CLOAKED for the test to return true.

easye (approved):

src/Cosi-BLS/new-transactions.lisp
markhdavid: It tests set-equal on two sets A, B. True if A is ```(:foo :bar)``` and B is ```(:bar :foo)```, thanks to sorting both lists. The first list is pre-sorted, saving a bit of time. The lists are tiny, so it's trivial time. Just didn't want to worry about it.

dbmcclain: why not just use (NULL (SET-DIFFERENCE a b)) ?

PR #263, opened 2018-06-28
Pt2 158153939 configure run minimal distributed

guitarvydas:
Not necessarily a PR, mainly examples for discussion.

In http://127.0.0.1:3140/ , create a new simulator, then examine output in "Output from Simulation".  This doesn't show *all* of the notes made during a run - only the notes that don't belong to the various nodes.

Click on Node API >> Get Status from Tracker.  This should show the rest of the notes, grouped by Node.  It also dumps the mempool and blockchain for each Node (after a run).  Look near the bottom of src/state-tracker.lisp for how this data was fetched and how it was formatted as strings.

The "system" notes show lots of left-over messages, like "Block validated ...".  These should belong to specific nodes, but don't at the moment - probably due to nested macros and magic swaps of specials (when switching nodes).  It is worth tracking these "problems" down only if we proceed with the CoSi simulator.  

The sim nodes (src/Cosi-BLS/cosi-construction.lisp) have been given a new slot - node-notestream - which is initialized to a string-stream for every sim Node.  The code in cosi-handlers.lisp has been changed to swap *notestream* when *current-node* is swapped (not all instances of this swapping were caught, as discussed above).  Most (if not all) output messages are sent through emotiq:note, and along with the swapping of *notestream*, this captures the notes from individual Nodes and keeps them in a string-stream slot of each individual Node, so that the REST interface can display notes that relate to each Node (without interleaving the notes).

The per-node notes are displayed in the Node API of the rest interface.  This might not be exactly what was intended.  The per-node info is created by src/state-tracker.lisp/get-all-info (a MAPC across all nodes, which can easily be broken up by examining the lisp code in get-all-info).  The call to get-all-info can be moved to any appropriate place in the REST interface.

Treat this as an example of what can be reported on a per-node basis and how to fetch this data.


easye:
@guitarvydas 

Gathering all the string messages from the simulator into a single REST resource <http://localhost:3140/simulator/0> looks fine. 

But what we really need is to turn this stream of messages into structured objects that we can deliver either via REST or push to various statistics resources (like Prometheus or Kafka) 

To this end, what we really need is to record more information in the state tracker, which then we can meaningfully transcribe into structured messages via JSON.

guitarvydas:
pausing...

PR #266, opened 2018-06-28
Lg svs8 pr

svspire:
Add stakes.conf generation and reading

### DIFF
--- a/src/gossip/config/generate.lisp
+++ b/src/gossip/config/generate.lisp
@@ -1,16 +1,28 @@
 (in-package :gossip/config)
 
+(defparameter *nominal-gossip-port* 65002 "Duplicate of same symbol in gossip package
+    to avoid package interdependencies")
+
+#+OBSOLETE
 (defparameter *aws-example* 
   '((:hostname "emq-01.aws.emotiq.ch" :ip "34.239.111.18" :port 65002)
-    (:hostname "emq-02.aws.emotiq.ch" :ip "52.54.224.84" :port 65002)
-    (:hostname "emq-03.aws.emotiq.ch" :ip "34.238.39.58" :port 65002)))
+    (:hostname "emq-02.aws.emotiq.ch" :ip "52.54.224.84"  :port 65002)
+    (:hostname "emq-03.aws.emotiq.ch" :ip "34.238.39.58"  :port 65002)))
 
-(defun generate (records &key (root #p"/var/tmp/conf/"))
+(defparameter *aws-example* 
+  '((:hostname "zt-emq-01.zt.emotiq.ch" :ip "10.178.2.166")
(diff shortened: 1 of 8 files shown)


src/gossip/config/generate.lisp
```diff
@@ -1,16 +1,28 @@
 (in-package :gossip/config)
 
+(defparameter *nominal-gossip-port* 65002 "Duplicate of same symbol in gossip package
+    to avoid package interdependencies")
+
+#+OBSOLETE
```
easye, line 6: @svspire Don't mark things as `#+obsolete`.  Please just delete, or if you really want to comment it out, use `#+(or)`

src/gossip/config/generate.lisp
```diff
@@ -21,25 +33,33 @@ The ROOT defaults to '/var/tmp/conf'."
           (emotiq:note "~&Writing configuration to '~a'.~&" directory)
           (ensure-directories-exist directory)
           (write-local-machine-conf
-           (merge-pathnames "local-machine.conf" directory)
+           (merge-pathnames *machine-filename* directory)
```
easye, line 36: @svspire Why would you ever want to change the name of the file?

markhdavid:
I loaded it and rand the config stuff. Things seem to run and build OK.  I will need some quality time to with the config stuff.  I'm ready to approve if Shannon's ready.

markhdavid (approved):
I'm ready if Shannon's ready.

easye (approved):

src/gossip/config/generate.lisp
svspire: Why not? #+(or) means nothing to me; #+obsolete signals this code probably needs to be removed at some point but I'm leaving it here for reference until the new code is thoroughly tested.

easye:
The feature `obsolete` may be present, `#+(or)` will never evaluate to truth.

src/gossip/config/generate.lisp
svspire: Famous last words. DRY principle.

easye:
> Famous last words. DRY principle.

It would be DRY if were repeated in the source, but it is not as far as I can see.

src/gossip/config/generate.lisp
svspire: > The feature obsolete may be present, #+(or) will never evaluate to truth.

If the feature obsolete is present, it's only because somebody put it there. In which case I'd have to ask why.
I also use the feature #+IGNORE sometimes during development. Same answer.

src/gossip/config/generate.lisp
svspire: It appears once in generate.lisp and twice in read.lisp.

svspire:
Variables that end in "-path*" probably should be gotten rid of at some point, since we're creating those on the fly now.

PR #267, opened 2018-06-28
Dbm ready to roll with gossip p

dbmcclain:
Up for grand review... I think this is ready to roll on AWS... Does not yet compile because it lacks GOSSIP:MAKE-NODE, GOSSIP:INITIALIZE-NODE, and GOSSIP:GET-STAKES. Once those are in place this ought to run.

Removed all semblance of election hold-off periods, etc. No need for that since we validate calls for election messages, and we are in charge of the 1-shot beacon.


easye:
@svspire What's the status on providing the needed Gossip function symbols that @dbmcclain lists as missing to make this work?

easye (changes requested):
Compilation failing, see CI.
```
package linda]...........
Failed to Quickload dependencies for system `COSI-BLS-TESTS`:
Undefined function ACTORS::PARSE-RECV-CLAUSES called with arguments ((:TIMEOUT ACTORS::TIMEOUT :ON-TIMEOUT (FUNCALL ACTORS::FN))) .
Exit code: 2
```

svspire:
> @svspire What's the status on providing the needed Gossip function symbols that @dbmcclain lists as missing to make this work?

Those functions should all be present in this branch.

easye:
@dbmcclain Running `emotiq:main` on this branch under LispWorks/macOS gives 
```
No applicable methods for #<STANDARD-GENERIC-FUNCTION COSI-SIMGEN::NODE-LOCAL-EPOCH 4290022864> with args (NIL)
   [Condition of type SIMPLE-ERROR]

```

dbmcclain:
Okay, finally merged with dev and Shannon's newest code. Should be ready to roll, sans simulator

dbmcclain:
Made the necessary patches to the sim driver to get it to bark. Looks like it is running now. But only you can tell me if it runs correctly. I inspected the `cosi-simgen:*top-node*` at the finish and the mempool was empty and the blockchain had several blocks in it.

dbmcclain:
RE: Missing NODE-LOCAL-EPOCH from emotiq:main ?? It is in there and defined as a slot accessor on Cosi Node structures. 

But I see a whole bunch of code in main.lisp, function LAUNCH-ROUTER, that will surely fail. Some messages no longer have responder methods (moved to ancient-cosi-code.lisp), and most of the message sends supply no args, which will surely bomb.

PR #269, opened 2018-06-29
New block slots for merkle roots and witness/stakes, cleanups continue.

markhdavid:
- added a slot to eblock class for a table (a-list) giving
  corresponding node public keys their stake amounts

- merkleize witness data and input scripts: added slots to eblock
  class for new merkle root hashes for witness data, and added these
  hashes to block creation, and added checks in block validation on
  non-leader nodes

- get rid of transaction contexts, an idea whose time came and went

- document and use more of account-addresses= for addresses (public
  key hashes)

- clean up doc on tx-ids=

- clean up doc on do-transactions

- wallet API fixups:
  - fix and change UTXO-P to make sense and document it reasonably
  - get-balance now works; fix up its doc, add a useful second value
  - package exports: utxo-p, get-utxos-per-account, get-balance

- add some convenience debug/repl util functions to-txid, find-tx
  - package exports: to-txid, find-tx

### DIFF
--- a/src/Cosi-BLS/block.lisp
+++ b/src/Cosi-BLS/block.lisp
@@ -51,9 +51,28 @@
     :documentation
     "Sequence of public keys of validators 1:1 w/signature-bitmap slot.")
 
+   (witnesses-and-stakes-table
+    :accessor block-witnesses-and-stakes-table
+    :documentation
+    "An a-list of the form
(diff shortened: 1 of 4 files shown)


markhdavid:
saw a bug, withdrawing this.

PR #275, opened 2018-07-02
Create-genesis-block takes witnesses-and-stakes.

easye:
Stored on block-witnesses-and-stakes slot of etable.
Block-witnesses-and-stakes-table renamed block-witnesses-and-stakes.
Symbol block-witnesses-and-stakes now exported from cosi/proofs.


markhdavid:
Was superseded by #277

PR #278, opened 2018-07-02
Emotiq config

easye:
* Introduce EMOTIQ/CONFIG to unify abstracted configuration strategies.

```(emotiq/config:network/generate)```

* provides REST/Websocket stubs for all needed functions around transactions

```(emotiq:start)```

Browse <http://localhost:3140>.

### CODE REVIEW

File: src/config.lisp [hunk 1]
```diff
@@ -0,0 +1,122 @@
...
+             (getf node :hostname))
+            (ip
+             (getf node :ip)))
+        (push (cons :hostname hostname)
+              configuration)
+        (push (cons :ip ip)
+              configuration)
+        (when settings-key-value-alist
+          (loop :for (key . value)
+             :in settings-key-value-alist
```
markhdavid: why is formatting inconsistent here?

src/config.lisp
```diff
@@ -0,0 +1,122 @@
...
+             :doing (push `(,key . ,value) configuration)))
+        (push `(:public
+                . ,(random 100)) ;; FIXME
+              configuration)
```
markhdavid: Why not stick to (cons x y) for `(,x . ,y) for this whole function for consistency.  The (cons x y) approach seems nice, since you can avoid the often confusing backquote and comma syntax, as well as not having to explicitly type and have to read the little tiny dot (.).

src/config.lisp
```diff
@@ -0,0 +1,122 @@
...
+      (:gossip-server
+       . :true)
+      (:gossip-server-port
+       . 65002)
+      (:genesis-block
+       . "emotiq-genesis-block.json")))
+
+(defun generated-directory (configuration)
+  (let ((host (alexandria:assoc-value configuration :host))
+        (ip (alexandria:assoc-value configuration :ip))
```
markhdavid: (1) why do you need to use alexandria:assoc-value? just to have setf'able assoc?  I do not even see where you're setf'ing the assoc. And even if you were, you can just setf the cdr of the result of assoc.
(2) instead of splattering a totally unabstracted advertisement of what your data structure is across the entire module, and maybe beyond, please abstract the access to one place: get-configuration-value or something like that

src/gossip/config/t/generate-network.lisp
```diff
@@ -1,42 +0,0 @@
...
-                    :port 443)
-         (:hostname "emq-05.aws.emotiq.ch"
-                    :ip "34.217.39.58"
-                    :port 22))))
-  (ensure-directories-exist root)
-  (prove:ok (gossip/config:generate network :root root)
-            (format nil "Generating test network configuration in~&~t~a~& for ~&~s~&…"
-                    root network))
-  (let* ((hostname
-          (getf (alexandria:random-elt gossip/config:*aws-example*)
```
markhdavid, line 26: No (1) you do not need to bring in alexandria:random-elt, and (2) you should not have actual randomness in a unit test, please.  
Generally, we want tests to produce the same exact results every time.
There's a way to make tests repeatable using random state, but who knows if that's a feature of alexandria blah blah. Who wants to know? I do not. Simpler is better.

src/genesis.lisp
```diff
@@ -0,0 +1,19 @@
...
+                                       root
+                                       (force nil))
+  (let ((genesis-block-path (merge-pathnames
+                             (alexandria:assoc-value  configuration
+                                                      :genesis-block)
+                             root)))
+    (unless (or force
+                (probe-file genesis-block-path))
+      (let ((genesis-block
+             (cosi/proofs:create-genesis-block
```
markhdavid, line 13: This is using an out-of-date call signature. This function takes a second arg now. This branch needs rebasing with dev.

src/gossip/config/generate.lisp
```diff
@@ -3,63 +3,43 @@
...
-           (merge-pathnames *hosts-filename* directory)
-           key-records)
-          (write-keypairs-conf
-           (merge-pathnames *keypairs-filename* directory)
-           key-records)
-          (write-stakes-conf
-           (merge-pathnames *stakes-filename* directory)
-           stakes)
-          (push directory directories))))
+        (let ((path (make-pathname :directory (append (pathname-directory root)
```
markhdavid: Pathname-directory is not guaranteed to return a list.
```
CLHS: X3J13 voted in June 1989 (PATHNAME-SUBDIRECTORY-LIST)   to specify that pathname-directory always returns nil, :unspecific, or a list
```

src/model/dictionary.lisp
```diff
@@ -0,0 +1,12 @@
+(in-package :model/wallet)
+
+(defun enumerate-dictionaries ()
+  "List all the available code dictionary by ISO code"
+  `("en"))
+
+(defun get-dictionary (iso-code)
+  "Get the 2048 word list for language with ISO-CODE"
+  (declare (ignore iso-code))
+  `(:|dictionary| ,(cosi-keying::import-wordlist "english.txt")))
```
markhdavid, line 10: Why is an unexported symbol being used here?

markhdavid (changes requested):
See comments.

src/config.lisp
easye: 'cuz that's how SLIME indents things by default.

src/config.lisp
easye: I find 
```
(assoc-value alist key)
```
to be more readable than the nesting
``` 
(cdr (assoc key alist))
```

src/config.lisp
markhdavid: Not for me. I get
```

          (loop :for (key . value)
                  :in settings-key-value-alist
                :doing (push `(,key . ,value) configuration))
```

src/config.lisp
markhdavid: Neither one is winning a beauty contest. Abstract it: (get-configuration-value ...)

src/genesis.lisp
easye: Addressed via merging with `dev`.

src/gossip/config/generate.lisp
easye: This code has been removed, as `gossip/config` no longer generates networks, but rather information for a node.

src/model/dictionary.lisp
easye: No public symbol is exported from `cosi-keying` that can be used.

The code list dictionary should be broken out of `cosi-keying`, and a proper API constructed in the future.

src/gossip/config/t/generate-network.lisp
easye: The code has been removed, as testing the network generation is now being done in `emotiq/config`

easye:
@markhdavid @svspire I have addressed outstanding comments, and the configuration code now seems like a plausible candidate for consolidation in `dev` (where it is needed for constructing test networks). 

Please review this pull request for approval.

markhdavid (approved):
I think things that needed changing were addressed, and other things were meant as work-towards type feedback, so not needed right now. Approved.

markhdavid:
Ugh, just noticed, "all checks have failed". Well, I'll plead, "CI diagnostics are being reviewed".

easye:
@markhdavid Fixed the CI failure with e5b4dcf 

Very odd problem with ASDF dependencies shadowing that I don't fully understand.

PR #286, opened 2018-07-04
User/developer-friendly transaction printer

markhdavid:


easye (approved):
Nice output here.  

Any reason we don't use a different character than  `#\!` to differentiate between spend and collect transactions?

markhdavid:
> Nice output here.

Thanks.

> Any reason we don't use a different character than #\! to differentiate between spend and collect transactions?

I tried a bunch, weighted extra good for being ASCII and for being shorter (1 or 2 characters), and giving some indication something was up. (Null is still in the running, near the top, by the way: it's what git does.) This was what I came up with. Worth experimenting, shooting around ideas again at some point.

> Any reason we don't use a different character than #\! to differentiate between **spend and collect** transactions?

Oh sorry I just justified the particular character I chose, not the question. Anyhow: good question. I tried, and I found it a bit confusing and also unnecessary because the address is already distinguished by being one longer in length and by always starting with one or another well-known character (for Emotiq system at least).

PR #287, opened 2018-07-04

easye:
Ensure that distinct keys are generated for stakes

Repackage cosi-bls with ASDF as cosi.asd

Document the stakes a little higher.

Addresses problems referencing EMOTIQ/CONFIG.

Correct PROVE:PLAN join point in test annotations.

Make use of alternate Cosi packaging in <file:src/Cosi-BLS/cosi.asd>
to resolve ASDF dependencies.  TODO: refactor everything that depends
on Cosi?  Too much work for now, just use as is.

### DIFF
--- a/src/Cosi-BLS/cosi-bls.asd
+++ b/src/Cosi-BLS/cosi-bls.asd
@@ -31,6 +31,7 @@ THE SOFTWARE.
   :depends-on (emotiq/logging
                ironclad
                actors
+               emotiq/config/stakes
                core-crypto
                crypto-pairings
                lisp-object-encoder
--- a/src/Cosi-BLS/cosi.asd
+++ b/src/Cosi-BLS/cosi.asd
@@ -1,0 +1,80 @@
+(defsystem cosi
+  :depends-on nil)
+
+(defsystem cosi/package
+  :depends-on (actors
+               core-crypto
+               ads-clos
+               crypto-pairings
+               emotiq/logging
+               ironclad
+               lisp-object-encoder
+               useful-macros
+               usocket
+               trivial-garbage
+               gossip)
+  :components ((:module package
+                        :pathname "./"
+                        :components ((:file "package")))))
+
+(defsystem cosi/bls/dependencies
+  :depends-on (actors
+               core-crypto
+               ads-clos
+               crypto-pairings
+               emotiq/logging
+               ironclad
+               lisp-object-encoder
+               useful-macros
+               usocket
+               trivial-garbage
+               gossip)
+  :components nil)
+
+(defsystem cosi/bls
+  :depends-on (cosi/package
+               cosi/bls/dependencies)
+  :components (#+ccl
+               (:module clozure
+                        :pathname "./"
+                        :components ((:file "clozure")))
+               (:module source
+                        :pathname "./"
+                        :components ((:file "cosi-blkdef")
+                                     (:file "cosi-keying")
+                                     (:file "cosi-netw-xlat")))))
+
+(defsystem cosi/simgen
+  :depends-on (cosi/package
+               cosi/proofs
+               cosi/transactions
+               cosi/proofs/newtx
+               emotiq/config/stakes)
+  :components ((:file "cosi-construction")
+               (:file "cosi-handlers")
+               (:file "mvp-election-beacon")))
+
+(defsystem cosi/transactions
+  :depends-on (cosi/bls
+               cosi/proofs/range)
+  :components ((:file "transaction")))
+
+
+(defsystem cosi/proofs/range
+  :depends-on (cosi/package)
+  :components ((:file "range-proofs")))
+
+
+(defsystem cosi/proofs
+  :depends-on (cosi/package)
+  :components ((:file "address")
+               (:file "block")))
+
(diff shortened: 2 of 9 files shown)


markhdavid (changes requested):
When this was tried, (emotiq/config:get-stakes) returns a list, but each element is in the wrong format. It's a plist of host and key data with not stake info. The element should be of the form key amount.  I notice the unit test only tests for non-nil result, not testing any of the results. Please fix and also extend the unit test.
*****
Update: the above pertained to the branch of this PR prior to latest commit. After I updated, there was still a problem: I regenerated the config files. Then I found:
(gossip:get-stakes) => nil, and (emotiq/config:get-stakes) is undefined.
So, still rejecting the PR on this basis at least.  I have other comments

markhdavid:
I also do not want this to go in with such a radical clone and modify change on a major system representing most of our code.  We have an entirely new defsystem and the old defsystems still in the source tree.  This is too chaotic. We need a simpler change.

easye:
@markhdavid The `(emotiq/config:get-stakes)` call works for me.

easye:
@markhdavid 
> I also do not want this to go in with such a radical clone and modify change on a major system representing most of our code. We have an entirely new defsystem and the old defsystems still in the source tree. This is too chaotic. We need a simpler change.

I have not removed the old ASDF definition for `cosi-bls`, it continues to function as before.  What sort of "simpler change" would you propose?

The definitions in <https://github.com/emotiq/emotiq/blob/emotiq-config/src/Cosi-BLS/cosi.asd> are not currently used in the main system load sequence.  I agree that substituting a new ASDF definition wholesale for `Cosi-BLS` would be too many changes.  Writing `cosi.asd` to separate out the dependencies helped me to understand and solve part of the circular ASDF dependencies problems by adding the ASDF dependency`cosi/proofs/newtx`  to a subsystem on `emotiq/config` avoiding the need to have `emotiq/config` depend on all of `cosi-bls`.

markhdavid:
Sorry, I was mistaken. The call works. This (emotiq/config:get-stakes) gets a result that's more like the correct format, but it's not the same format. However, the format has changed incompatibly, from (private-key stake) to (private-key . stake).

markhdavid:
I would prefer an incremental change to existing files, but I do not know what that change should be.  Could you give an explanation of what is the basic change in the system this is accomplishing? Why can't it be done with a change to existing defsystem?

easye:
@markhdavid 
>  would prefer an incremental change to existing files,

This pull request does make an incremental change to existing ASDF definitions, by allowing `cosi-bls` to include the parts of the configuration system it needs.  

To repeat myself, the  `cosi.asd` is *not* currently used except as needed by `emotiq/config`.  It also serves as a demonstration to the developer (me!) of what the `cosi-bls` dependencies are by allowing their compilation to tighter dependency semantics.  This has been useful as a fallback while I figure out the correct way to incrementally change existing files, which is what this pull request does.

easye:
> (emotiq/config:get-stakes) gets a result that's more like the correct format, but it's not the same format. However, the format has changed incompatibly, from (private-key stake) to (private-key . stake).

Sorry about that.  The documentation you wrote for `normalize-witnesses-and-stakes` must be incorrect:

```
(defun normalize-witnesses-and-stakes (witnesses-and-stakes)
  "Given an a-list of the form

     ( (general-public-key . (emtq-amount-in-subunits))* )

```

Sigh.  I had to change David's API calls as well.

@markhdavid Do you wish me to change back the stakes back to a list of lists?

markhdavid:
The new format of stakes would not work with setup-emergency-call-for-new-election in src/Cosi-BLS/mvp-election-beacon.lisp. The genesis block also has this information stored in slot witnesses-and-stakes (intended for future use) in the same format, which is not the format documented and specified.

easye:
> The new format of stakes would not work with setup-emergency-call-for-new-election in src/Cosi-BLS/mvp-election-beacon.lisp

It should, as I changed that code.

markhdavid:
No, I do not think it's incorrect. Note that (key . (stake)) is the same as (key stake). Sorry if the notation was confusing, but it is classic Lisp dot notation.

markhdavid:
The code in my branch looks like it wants (key stake) format:
```
        (setf (node-stake node) ;; probably never used elsewhere...
              (second (assoc (node-pkey node) (get-witness-list)
                             :test 'int=))))
```
from the code on this branch in mvp-election-beacon.lisp

markhdavid:
We reviewed the issue of adding the defsystem. We agreed in the medium term the config system can and should be refactored so that a part can be depended on by the core system to serve config info, and there can be a part that depends on the core system that make direct calls to core system functions like create-genesis-block.  However, I feel for now we can live with having the whole system stay as it was, that is, not dependent on core (anything system cosi in its name), so for now for generating config files, users or processes will be required to load cosi by some mechanism not provided here (e.g., by hand, or by scripts done by devops). For now, the call to a cosi function can simply be a funcall to a hook function or even an intern of string. There's only one case (call to create-genesis-block), and this is medium-term temporary until the above mentioned refactoring, so it does not matter if it's slightly inelegant.

easye:
Superseded by #293

PR #288, opened 2018-07-04
write abstractions on top of usocket able workaround some usocket bugs

markhdavid:
I think this will make everything a little better. It's for the case on MacOS with LispWorks where you have a localhost that gets resolved to an ipv6 address. If 
```
 (comm:get-host-entry "localhost" :fields '(:addresses))
=> (#<COMM:IPV6-ADDRESS ::1 4020187E73> 2130706433)
```
Gossip-startup was getting an error on LispWorks on MacOS.  This makes a bit of a layer on top of those usocket functions, with a workaround. This fixes the error that occurs at startup. However, it apparently only affected my machine per reports so far. But I kind of understand why that's possible now.

The "remove" loop in gossip startup is way simpler and easier for me to understand. Hope Shannon finds it an improvement.

### DIFF
--- a/src/gossip/gossip-startup.lisp
+++ b/src/gossip/gossip-startup.lisp
@@ -14,7 +14,7 @@
 (defun process-eripa-value (ev)
   (setf *eripa* (if (eq :deduce ev)
                     (eripa)
-                    (usocket::host-to-vector-quad ev))))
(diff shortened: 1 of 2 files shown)


easye:
`gossip-tests` failing 5 tests but not because of this commit.

easye (approved):

PR #293, opened 2018-07-05

easye:
Ensure that distinct keys are generated.

Use funcall instead of ASDF dependency to generate configurations.

Correct PROVE:PLAN join point in test annotations.

Fix election requests (dbm)

GENESIS/CREATE returns the created block as well as serializing it

Remove references to stakes from GOSSIP

### DIFF
--- a/src/Cosi-BLS/cosi-construction.lisp
+++ b/src/Cosi-BLS/cosi-construction.lisp
@@ -200,7 +200,12 @@ THE SOFTWARE.
 ;; new for Gossip support, and conceptually cleaner...
 (defmethod initialize-instance :around ((node node) &key &allow-other-keys)
   (setf (node-self node) (make-node-dispatcher node))
-  (call-next-method))
+  (call-next-method)
+  (let ((genesis-block (emotiq/config:get-genesis-block)))
+          (push genesis-block (node-blockchain node))
+          (setf (gethash (cosi/proofs:hash-block genesis-block)
+                         (node-blockchain-tbl node))
+                genesis-block)))
 
 ;; --------------------------------------------------------------
 
--- a/src/Cosi-BLS/mvp-election-beacon.lisp
+++ b/src/Cosi-BLS/mvp-election-beacon.lisp
@@ -292,7 +292,7 @@ based on their relative stake"
       ;; *local-epoch* will also not have
       ;; changed
       (unless (get-witness-list)
-        (set-nodes (gossip:get-stakes))  ;; <<--- THIS NEEDS TO CHANGE TO MARK D's NOTIONS
+        (set-nodes (emotiq/config:get-stakes))
         (setf (node-stake node) ;; probably never used elsewhere...
               (second (assoc (node-pkey node) (get-witness-list)
                              :test 'int=))))
@@ -331,7 +331,7 @@ based on their relative stake"
 
 (defun make-signed-call-for-election-message (pkey epoch skey)
   (let ((skel  (make-call-for-election-message-skeleton pkey epoch)))
-    (um:conc1 skel (pbc:sign-hash (hash/256 skel) skey))))
+    (um:append1 skel (pbc:sign-hash (hash/256 skel) skey))))
 
 (defun validate-call-for-election-message (pkey epoch sig)
   (and (= epoch *local-epoch*)        ;; talking about current epoch? not late arrival?
--- a/src/genesis.lisp
+++ b/src/genesis.lisp
@@ -1,29 +1,37 @@
 (in-package :emotiq/config)
 
 (defun genesis/create (configuration &key
-                                       root
+                                       (directory (emotiq/fs:tmp/))
                                        (force nil))
-  (let ((genesis-block-path (merge-pathnames
-                             (alexandria:assoc-value configuration
-                                                     :genesis-block-file)
-                             root)))
-    (when (or force
-              (not (probe-file genesis-block-path)))
-      (let ((genesis-block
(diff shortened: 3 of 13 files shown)


src/genesis.lisp
```diff
@@ -1,29 +1,37 @@
...
-              (not (probe-file genesis-block-path)))
-      (let ((genesis-block
-             (cosi/proofs:create-genesis-block
-              (alexandria:assoc-value configuration :address-for-coins)
-              (get-stakes))))
-        (with-open-file (o genesis-block-path
-                           :direction :output
-                           :if-exists :supersede)
-          (cl-json:encode-json genesis-block o))
+  (let ((create-genesis-block (uiop:find-symbol* :create-genesis-block :cosi/proofs)))
```
markhdavid, line 6: The package may not exist. Check that first. Could wrap what's there with ```(and (find-package :cosi/proofs) ...)```.

markhdavid (changes requested):
Bug around the check for fboundp. See inline comments.

src/Cosi-BLS/cosi-construction.lisp
```diff
@@ -200,7 +200,12 @@ THE SOFTWARE.
 ;; new for Gossip support, and conceptually cleaner...
 (defmethod initialize-instance :around ((node node) &key &allow-other-keys)
   (setf (node-self node) (make-node-dispatcher node))
-  (call-next-method))
+  (call-next-method)
```
markhdavid, line 203: This change, introducing the genesis block, is in a branch, and I want to do it a different way. It was not in the earlier version of the PR or of this branch, from yesterday, that I reviewed. I would prefer to remove it, and I will add something equivalent but different subsequently. The sequencing works nicely, since the election code in cosi still does not really depend on this.  (Also, by the way, this change was not mentioned in the commit log message, which it should have been if it were going in.)

markhdavid (changes requested):

src/Cosi-BLS/mvp-election-beacon.lisp
```diff
@@ -331,7 +331,7 @@ based on their relative stake"
 
 (defun make-signed-call-for-election-message (pkey epoch skey)
   (let ((skel  (make-call-for-election-message-skeleton pkey epoch)))
-    (um:conc1 skel (pbc:sign-hash (hash/256 skel) skey))))
+    (um:append1 skel (pbc:sign-hash (hash/256 skel) skey))))
```
markhdavid, line 334: What is this change, which is vaguely explained as ```Fix election requests (dbm)``` in the commit log message, doing here a commit for the config system.  It's by a different developer for a different system, by all indications. Therefore, it does not belong in this PR.  The commit should be redone without if they were done together. I have not checked the history.

markhdavid (changes requested):

src/Cosi-BLS/mvp-election-beacon.lisp
easye: > What is this change

This branch was intended to show that `emotiq/config` can be used to run a chain.  As such it contains code to add the genesis block.  

The change to fix election requests <https://github.com/emotiq/emotiq/pull/289/commits/00c71d3956c5cfe88994de768caa8c082bdf88bf> was noted by @dbmcclain as necessary for multiple elections to succeed.

src/genesis.lisp
easye: Grumble.  I thought that `uiop:find-symbol*` returned nil if the package didn't exist, but it doesn't.  Will fix.

src/Cosi-BLS/mvp-election-beacon.lisp
markhdavid: We just want small, coherent commits. This was a replacement for a previous PR, and it should address the issues that were there, which it largely has, not add new features or fix other bugs opportunistically.  If David's fix is needed, please do it as a separate PR, or have David do it.  The genesis one would be fine to do as a followup separate PR, but as I've told you I have slightly different commit/PR waiting for that, so I might as well do it.

markhdavid: Oh I just noticed it looks like the above is actually a currently approved PR: Fixup: UM:CONC1 should have been UM:APPEND1 for non-destructive behavior #289 . Accordingtly, it should be put on dev, and this branch should be rebased on top of it.

easye:
> On Jul 6, 2018, at 07:43, markhdavid <notifications@github.com> wrote:
> 
> @markhdavid commented on this pull request.
> 
> In src/Cosi-BLS/mvp-election-beacon.lisp:
> 
> > @@ -331,7 +331,7 @@ based on their relative stake"
>  
>  (defun make-signed-call-for-election-message (pkey epoch skey)
>    (let ((skel  (make-call-for-election-message-skeleton pkey epoch)))
> -    (um:conc1 skel (pbc:sign-hash (hash/256 skel) skey))))
> +    (um:append1 skel (pbc:sign-hash (hash/256 skel) skey))))
> 
> We just want small, coherent commits. This was a replacement for a previous PR, and it should address the issues that were there, which it largely has, not add new features or fix other bugs opportunistically. If David's fix is needed, please do it as a separate PR, or have David do it. The genesis one would be fine to do as a followup separate PR, but as I've told you I have slightly different commit/PR waiting for that, so I might as well do it.

This pull request was about making and demonstrating  `emotiq/config`  was usable, which needed the interfaces between the systems as demonstrated as working.  

As such, it was a small, coherent commit that increased the progress towards getting towards a working chain.

PR #298, opened 2018-07-06

easye:
Ensure that distinct keys are generated.

Use funcall instead of ASDF dependency to generate configurations.

GENESIS/CREATE returns the created block as well as serializing it.

Remove references to stakes from GOSSIP.

Subsumes #287


svspire:
I spent all day yesterday testing and exploring this code. I can't approve the PR yet because the dependency on COSI-BLS means Ansible is not able to create genesis block files, and thus not able to copy them to AWS nodes.

Potential solution: Given that we have stakes.conf files, there's no need for #'network/generate to make genesis blocks. Instead, #'main could write it to emotiq/var/etc/ automatically if it doesn't already exist. It would be easy to tell Ansible not to try to copy the genesis block files. Any downsides to this idea?

Ultimately of course, we need to get rid of stakes.conf and keypairs.conf files, and will have to find a better solution. This is just for MVP.

easye:
@svspire No need for such complications: one just has to rewrite the Ansible configuration scripts to explicitly load `cosi-bls`.   See <https://github.com/emotiq/devops-ansible/pull/15> for the code.

svspire (approved):

PR #301, opened 2018-07-09
Overhaul test infrastructure to return nil status on failure

easye:
*N.b* One needs to update the `emotiq` Quicklisp distribution to obtain the `asdf-test-harness` dependency. 
q.v. 
```lisp
CL-USER> (ql-dist:all-dists)
(#<DIST emotiq 20180525154925> #<DIST quicklisp 2018-01-31>)
CL-USER> (ql:update-dist (first (ql-dist:all-dists)))
```
Working invocations of `etc/test-harness.bash` for the developer and the CI.

Depends on ASDF-TEST-HARNESS from the Emotiq Quicklisp distribution

Consistently name test packages and directories 'test'.

`etc/test-harness.bash` works with new system under CCL.  We need to
revisit the script invocation for developer, but the CI should work
for all the CL implementations we test as it has a script which
normalizes the command line options for the `--load` and `--eval`
options.  We should look at how to use this interface for the interactive developer.


echupriyanov (approved):
Fine as a starting point

PR #303, opened 2018-07-11
Use configuration of REST and Websocket listen addresses

easye:
@echupriyanov This code actually uses the configuration for the REST/Websocket routines, which will be necessary to avoid port collisions on the local machine running more than one process.


echupriyanov (approved):
Good starting point

PR #304, opened 2018-07-11
Hello message2 pr

svspire:
We're now back to the functionality we had two weeks ago before the configuration architecture changed. AWS nodes are talking to each other and elections are being called. And the new hello message works now.

### DIFF
--- a/src/filesystem.lisp
+++ b/src/filesystem.lisp
@@ -11,36 +11,53 @@
 ;;;; time, we have to somehow "walk" their reference to copy them for
 ;;;; a "resource bundle" that can then be distributed.
 
+(defparameter *subpath* nil "Directory subpath to be added to each of the below. Must end with /. Useful for testing.")
+
 (defun var/log/ ()
   "Absolute cl:pathname of the directory to persist logs and traces of system activity."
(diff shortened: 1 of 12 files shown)


dbmcclain:
Well, at first it bombed horribly with a segfault.... Then I realized that I had my new version of the PBC library installed from a few days ago. No wonder... 

So I recompiled the version of PBC you are using, installed the lib, and your code now loads up okay. Could not run my Randhound because your branch is too far behind. But I did run Emotiq/Sim and it appears to run okay.

There is one error message on the first load up of the code. It started happening yesterday, so probably unrelated to this PR:

EMOTIQ/CONFIG::CONFIGURATION which is referenced by EMOTIQ/CONFIG:GET-GENESIS-BLOCK

dbmcclain (approved):

svspire:
Sounds like we have some merging issues with your latest code. We can coordinate tommorrow. I did rebase from dev before I submitted the PR, so I'm going to go ahead and merge this now.

dbmcclain:
was not a merge issue… I did a checkout on your PR branch, so none of my session should have been present. I also deleted the cache and started a fresh Lisp image before testing.

PR #305, opened 2018-07-12
Add :config-subpath option to main so now you can say

svspire:
Add :config-subpath option to main so now you can say
`(emotiq:main :config-subpath "nodeN/")`
to enable running multiple emotiq nodes on localhost.

If you supply this arg, config files (and others defined in `filesystem.lisp`) will have the given subpath appended to them e.g.

instead of the default `../var/etc/`
config files will be found in `../var/etc/nodeN/`.

Other functions like `tmp/` etc are affected similarly.

### DIFF
--- a/src/startup.lisp
+++ b/src/startup.lisp
@@ -23,9 +23,17 @@
 
 (in-package "EMOTIQ")
 
+(defun directorify (str)
+  "Ensure str doesn't start with #\/ but it does end with one"
+  (let ((slash #.(string #\/)))
+    (setf str (string-trim slash str))
+    (setf str (concatenate 'string str slash))))
+
 ;; "Entry Point" for development - does nothing, just load and go
-(defun main (&optional how-started-message?)
+(defun main (&key config-subpath how-started-message?)
   (setf cosi-simgen::*use-real-gossip* t) ;; make sure code knows 
+  (when config-subpath
+    (setf emotiq/fs:*subpath* (directorify config-subpath)))
   (message-running-state how-started-message?)
   ;; Create a default wallet on disk if one doesn't already exist
   (emotiq/wallet:create-wallet)
--- a/src/config.lisp
+++ b/src/config.lisp
@@ -14,7 +14,7 @@
     (:hostname "zt-emq-03.zt.emotiq.ch"
      :ip "10.178.15.71")))
 
-(defparameter *emotiq-conf*
+(defun emotiq-conf ()
   (make-pathname :defaults (emotiq/fs:etc/)
                  :name "emotiq-conf"
                  :type "json"))
@@ -123,7 +123,7 @@
    :private (alexandria:assoc-value configuration :private)
    :key-records key-records)
   (with-open-file (o (make-pathname :defaults directory
-                                    :name (pathname-name *emotiq-conf*)
+                                    :name (pathname-name (emotiq-conf))
                                     :type "json")
                      :if-exists :supersede
                      :direction :output)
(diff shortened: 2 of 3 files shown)


easye (changes requested):
@svspire 

It should be the "other way around", namely  `(emotiq/fs:etc)` should be rooted in different directories for a different configuration, not have sub configurations within itself.  

I will create a different patch to show you what I mean.

I really want to continue to use the functional interface.

echupriyanov:
For my purposes Shannon's approach is fine also

echupriyanov (changes requested):
Well, looks like solution is incomplete:
```lisp
? (emotiq:main :config-subpath "node1/")

Running interactively in development
with args [(ccl)]
2018-07-12 16:27:22 Serializing wallet to /Users/eric/Library/Application Support/Emotiq/node1/wallet/node1/My%20Wallet/emotiq.wallet
2018-07-12 16:27:22 No configuration able to be read from '/Users/eric/Copoka/Emotiq/emotiq/var/etc/emotiq-conf.json'
2018-07-12 16:27:22 No configuration able to be read from '/Users/eric/Copoka/Emotiq/emotiq/var/etc/emotiq-conf.json'
2018-07-12 16:27:22 running start-tracker #<SYSTEM-STATE #x30200333B82D> #<ACTOR #x30200333B34D>
2018-07-12 16:27:22 Starting gossip configuration for testnet…
2018-07-12 16:27:22 Gossip init finished.
2018-07-12 16:27:22 Sleeping for 7 seconds.
#<ACTORS:ACTOR #x30200336B2CD>
? 
```

svspire:
Failure was because `*emotiq-conf*` was defined as a variable rather than a function. Fixed now.

echupriyanov (approved):

svspire:
@easye > ... should be rooted in different directories for a different configuration, not have sub configurations within itself

I don't see how this can work, given that `emotiq/user/root/`, `var/log/`, `etc/`, `tmp/`, and `emotiq/wallet/` exist at very different places on the file system, and in some cases depend on the particulars of the operating system. A consistent subdirectory name is the only way to keep all the files associated with a given node identifiable with each other.

> I really want to continue to use the functional interface.

What does this mean?
If you want to call `etc/` and the other filesystem functions with a parameter, that's okay by me, but every caller is going to have to be changed to use that parameter. And that parameter is usually going to be effectively a constant. I suppose we can make it an optional parameter to every function and have it default to `*subpath*`.

echupriyanov:
Looks like now it does what I need.

easye:
>emotiq/user/root/, var/log/, etc/, tmp/, and emotiq/wallet/ exist at very different places on the file system, and in some cases depend on the particulars of the operating system. A consistent subdirectory name is the only way to keep all the files associated with a given node identifiable with each other.

`var/log/`, `etc/` are rooted under `var`.  These should be under the same root directory.  That root directory should change for a given configuration

`tmp` is for temporary writable storage, which is indeed system dependent.  

`emotiq/user/` is the start of rooting all these files.  `etc` should actually be under there, but because we essentially have two wallet mechanisms at this point I didn't root that under `etc` yet.

easye:
@svspire As I understand things, this has already been merged to `dev` under a different pull request.  If so, please close this pull request.

svspire:
> As I understand things, this has already been merged to dev under a different pull request. If so, please close this pull request.

No it hasn't.

easye (approved):
@svspire 
I remove myself from the review due to fundamental issues from how this is implemented, but "approve" this in the interest of being willing to accumulate unlimited technical debt of working code in order to get it working, as the former is currently free.

src/startup.lisp
```diff
@@ -23,9 +23,17 @@
 
 (in-package "EMOTIQ")
 
+(defun directorify (str)
```
easye, line 26: The right way to do this is to convert to a `cl:pathname` rather than relying on namestring semantics:  you can't open a file without a valid `cl:pathname`

PR #312, opened 2018-07-15
Test harness 20180715b

easye:
Better intergration of tests for both the CI and the developer

CI now actually uses the contents of <file:etc/all-tests.sexp> to
determine which tests should be run.

Changes the mechanism used in test harness so that LispWorks can be
invoked outside the CI.

ci/lisp-wrapper-location.bash provides a function to interpolate a
given name of Lisp implementation ("lispworks", "ccl" etc.) into an
actual path to invoke.

Invoking the test harness from the shell now reads the database of all
defined tests in <file:etc/all-tests.sexp> rather than using a default
value.


echupriyanov (approved):
`:EMOTIQ-TEST` is failing in CI

Besides that PR is fine

Issue #310, opened 2018-07-13
JSON-RPC 2.0 response is not valid


echupriyanov:
I'm trying to use Python to talk to Emotiq WS server.
I can open connection and send `ping` request using JSON-RPC 2.0, but response from the Emotiq fails the validation:
```
jsonschema.exceptions.ValidationError: {'jsonrpc': '2.0', 'id': 1, 'result': 'pong', 'error': None} is not valid under any of the given schemas

Failed validating 'oneOf' in schema:
    {'$schema': 'http://json-schema.org/draft-04/schema#',
     'definitions': {'common': {'not': {'description': 'cannot have result '
                                                       'and error at the '
                                                       'same time',
                                        'required': ['result', 'error']},
                                'properties': {'id': {'note': ['spec says '
                                                               'a number '
                                                               'which '
                                                               'should not '
                                                               'contain a '
                                                               'fractional '
                                                               'part',
                                                               'We choose '
                                                               'integer '
                                                               'here, but '
                                                               'this is '
                                                               'unenforceable '
                                                               'with some '
                                                               'languages'],
                                                      'type': ['string',
                                                               'integer',
                                                               'null']},
                                               'jsonrpc': {'enum': ['2.0']}},
                                'required': ['id', 'jsonrpc'],
                                'type': 'object'},
                     'error': {'allOf': [{'$ref': '#/definitions/common'},
                                         {'properties': {'error': {'properties': {'code': {'note': ['unenforceable '
                                                                                                    'in '
                                                                                                    'some '
                                                                                                    'languages'],
                                                                                           'type': 'integer'},
                                                                                  'data': {'description': 'optional, '
                                                                                                          'can '
                                                                                                          'be '
                                                                                                          'anything'},
                                                                                  'message': {'type': 'string'}},
                                                                   'required': ['code',
                                                                                'message'],
                                                                   'type': 'object'}},
                                          'required': ['error']}]},
                     'success': {'allOf': [{'$ref': '#/definitions/common'},
                                           {'required': ['result']}],
                                 'description': 'A success. The result '
                                                'member is then required '
                                                'and can be anything.'}},
     'description': 'A JSON RPC 2.0 response',
     'oneOf': [{'$ref': '#/definitions/success'},
               {'$ref': '#/definitions/error'},
               {'items': {'oneOf': [{'$ref': '#/definitions/success'},
                                    {'$ref': '#/definitions/error'}]},
                'type': 'array'}]}

On instance:
[...]

PR #316, opened 2018-07-16
Normalize missing JSON RPC 2.0 elements as per specification

easye:
Addresses <https://github.com/emotiq/emotiq/issues/310>


echupriyanov (approved):
Now it gives correct response both in successful and error cases. 👍

PR #317, opened 2018-07-16
Dbm fastmod

dbmcclain:
This PR is just to allow people to see what I'm up to... please don't merge just yet. Still writing unit tests


src/Crypto/core-crypto-tests.asd
```diff
@@ -0,0 +1,18 @@
+
```
easye: @dbmcclain The convention I am trying to follow with defining test systems is to use the singular `-test` (instead of `-tests` for the package and ASDF systems.  While this may seem very "nitpicky", it would be nice to all follow this convention, as I normalized all other test systems to this convention.

dbmcclain: okay FB

dbmcclain:
@easye I am currently stuck in limbo. CCL has a bad interaction with CFFI on MacOSX 10.13.6, and CCL is refusing to load a library that is present. I suspect, but cannot prove, that the problem arises because the PBC_Intf lib calls another dylib libPBC, which in turn calls another dylib libGMP. The other two libs that do get loaded successfully are standalone and call on no other libs for anything.

So... Either we statically link, and hope that solves the problem, or else we have to punt on CCL until the CFFI/CCL interaction gets solved. I have tracked the problem down to one function in cffi-mcl.lisp and I have told Shannon about it.

Meanwhile, as long as your Mac is old enough it seems to run properly.

So what to do about the pull requests???

(I have also noticed that GitLab is broken)

echupriyanov:
@dbmcclain Please, check the thread on Twist.

easye:
@dbmcclain I am looking at the problems with CCL now with Eugene. (joining us on Twist would probably be faster)

PR #318, opened 2018-07-17
Svs bc pr

svspire:
Incorporates MHD's latest changes, plus a few corrections. cosi-bls-test seems to work on both LW and CCL.


easye (changes requested):
Let's get this to the state that we have a test before merging to `dev`.  

@guitarvydas What do you think?

svspire:
Every commit in this PR is included in svs-newtx, so I'm closing this PR.

PR #322, opened 2018-07-18
Add test for simulator to keep it working

### LINKED ISSUE #321: Simulator is lacking a test
While we get the network transport working, it is important to keep the simulator working as it is currently the only source of "truth" about whether transactions work or not.

As such, it is important not to break the simulator.

easye:
While we get the network transport working, it is important to keep the simulator working as it is currently the only source of "truth" about whether transactions work or not.

Closes #321.

### DIFF
--- a/etc/all-tests.sexp
+++ b/etc/all-tests.sexp
@@ -2,5 +2,6 @@
  :gossip-test
  :crypto-pairings-test
  :emotiq-config-test
- :cosi-bls-test)
+ :cosi-bls-test
+ :emotiq-sim-test)
  
--- a/src/Cosi-BLS/cosi-bls.asd
+++ b/src/Cosi-BLS/cosi-bls.asd
@@ -31,6 +31,7 @@ THE SOFTWARE.
   :depends-on (emotiq/logging
                ironclad
                actors
+               emotiq/utilities
                emotiq/tracker
                emotiq/config
                core-crypto
--- a/src/Cosi-BLS/cosi-construction.lisp
+++ b/src/Cosi-BLS/cosi-construction.lisp
@@ -381,6 +381,9 @@ THE SOFTWARE.
                      :text txt)))
 
   (defmethod view-tree ((tree node) &key (layout :left-right))
+    (unless (emotiq:x11-display-p)
+      (emotiq:note "No X11 display available to view tree")
+      (return-from view-tree nil))
     (capi:contain
      (make-instance 'capi:graph-pane
                     :layout-function layout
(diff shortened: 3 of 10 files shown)


easye:
LispWorks fails because it tries to open the inspector: 
```
;; Created directory /builds/emotiq/emotiq/src/Cosi-BLS/config/
 | Execution error:
 | Failed to open the default X11 display { Unix environment variable DISPLAY not set }
 |
``` 

@echupriyanov Can we run something under CI that will just silently accept X11 display call requests?  I think there is such a thing as a "null X Server", but my memory is failing.  Nothing needs to be input on the X server or anything, just the call to open the connection and call the X11 render routines need to finish.

ks (approved):

easye:
> Can we run something under CI that will just silently accept X11 display call requests?  

I will code for the presence of the `DISPLAY` environment variable to a null code path for the display of the inspector.

echupriyanov:
@easye yes, definitely there is 'null X11'. I use it to create non-GUI LispWorks image in Linux.
Its in `ci/install-lispworks-linux.sh` file:
```
xvfb-run $PREFIX/lispworks-7-1-0-amd64-linux -build /tmp/resave.lisp
```

easye:
@echupriyanov Then why is the test for this pull request failing on the CI?  Is that null-X11 server not being used?

Answer:  we'd have to invoke the test harness under `xvfb-run`.  @echupriyanov too complicated, right?  As we'd have to install `xvfb-run` (and most of X11?) under MacOS.

echupriyanov:
@easye I use this null-X11 only during generation of `lwpro` image. You need to wrap your shell command with `xvfb-run` to make `$DISPLAY` to be available. I'll take a closer look at the code.

easye:
@echupriyanov Please don't prioritize the adding of `xfvb-run` right now (you have enough to do).  I will fix by a check for the existence of a `DISPLAY` environment variable.

Since the presence of X11 is not a requirement at all for production, we should fix this in the code rather than the test environment.  Cleaner, and more correct to what we want to test.

echupriyanov:
@easye yes, fixing that in code definitely way to go. anyway I checked following under Linux:
```
lisp=lwpro systems=:emotiq-sim-test xvfb-run etc/test-harness.bash 
```
And it worked ok

echupriyanov (changes requested):
Let's make sure all `:emotiq-sim-test` passed before merging

easye:
> Let's make sure all :emotiq-sim-test passed before merging

easye:
Tests pass now with checking for X11 display.  @echupriyanov I need you to get your review to "accepted" in order to pull into `dev`.

echupriyanov (approved):

echupriyanov:
@mark good to go

PR #323, opened 2018-07-18
Dbm fastmod ec

dbmcclain:


src/Cosi-BLS/range-proofs.lisp
```diff
@@ -693,421 +695,578 @@ nbits pts will become right vectors."
...
+
+#||#
+;; ---------------------------------------------------------------------------
+;;
+;; for 128 bit security |#
 #|
 ;; Constants obtained from:
 
-       (with-ed-curve :curve-1174
+       (with-ed-curve :curve1174
```
easye, line 990: I kinda like the style of using `#\-` `for-separating-words` better, but not necessary to change this now.

Your usage here is certainly consistent with how the C code names things, which doesn't have the ability to use `#\-" as such a separator.

src/Crypto/core-crypto.asd
```diff
@@ -37,16 +37,16 @@ THE SOFTWARE.
                 (:file "hash")
                 (:file "ctr-hash-drbg")
                 (:file "primes")
-                #+:COM.RAL (:file "crypto-le")
-                (:file "kdf")
-                (:file "gf-571")
+                #+(AND :MACOSX :COM.RAL) (:file "crypto-le")
```
easye: It would be nice to remove or start deprecating the use of the `:com.ral` for `cl:*features*`.  

The use of `:com.ral` needs to be explained to every new developer, and isn't clear what it means.  

Since you always have the `:com.ral` feature present when you develop, I think it quite likely that there will eventually be some kind of bug that bites us because you are the only one that uses that code path.

src/Crypto/core-crypto.asd
```diff
@@ -55,5 +55,48 @@ THE SOFTWARE.
                  "s-base64"
                  "emotiq"
                  "emotiq/delivery"
+                 "cffi"
+		 "core-crypto/libraries"
+                 "crypto-pairings/libraries"
                  ))
 
+(defsystem "core-crypto/libraries"
```
easye: Probably better to start a new hierarchy for the various external libraries that `core-crypto` needs.

Perhaps:

 "core-crypto/libraries" --> "core-crypto/lib/curve1174"
"crypto-pairings/libraries" --> "crypto-pairings/lib/pbc"

src/Crypto/crypto-pairings.asd
```diff
@@ -26,37 +26,14 @@ THE SOFTWARE.
   :description "crypto-pairings: bilinear pairings (PBC) functions"
   :version     "1.1.0"
   :author      "D.McClain <dbm@refined-audiometrics.com>"
-  :in-order-to ((test-op (test-op "crypto-pairings-test")))
+  :in-order-to ((test-op (test-op "crypto-pairings-tests")))
```
easye, line 29: Please use the singular for for packages and ASDF systems to be consistent:

"crypto-pairings-tests" --> ""crypto-pairings-test"

src/Crypto/core-crypto.asd
```diff
@@ -22,13 +22,13 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 |#
 
-(in-package :cl-user)
-
 (asdf:defsystem "core-crypto"
   :description "core-crypto: core cryptography functions"
   :version     "1.0"
   :author      "D.McClain <dbm@refined-audiometrics.com>"
+  :in-order-to ((test-op (test-op "core-crypto-tests")))
```
easye, line 29:
```
(asdf:test-system :core-crypto)
```
fails because this line should be 

```
:in-order-to ((test-op (test-op "core-crypto-test")))
```

easye (commented):
You should include the new tests in the test harness by adding the system you wish to test in `etc/all-tests.sexp`.

easye (commented):
If this pull request subsumes <https://github.com/emotiq/emotiq/pull/300/> as it seems to from a basic check of changes, then please close #300

easye (changes requested):
Please remove the "randhound-old" files and references.  While I am pretty tolerant of slop and technical debt accumulation, it is a "bridge to far" to have such things in a commit that introduces RandHound in the first place.  

Since RandHound isn't needed for the first MVP, we can take our time here.

src/Cosi-BLS/range-proofs.lisp
dbmcclain: My naming conventions follow the academic literature (or try to), so Curve1174 is its name in print, as is Ed3363, where I do separate as curve-Ed3363 in order to keep it clear in the code that it is an ECC curve.

easye: Fine then.

src/Crypto/core-crypto.asd
dbmcclain: yes started thinking about that last night as I went weeding... Perhaps time to put a clear separation between my other uses of the crypto libs

src/Crypto/core-crypto.asd
dbmcclain: good idea

src/Crypto/core-crypto.asd
easye: Again, not necessary for this pull request, but should be done sooner rather than later.

src/Crypto/crypto-pairings.asd
dbmcclain: yes, I thought I did that yesterday. I like your consistent convention. Looks like I need to go back and double check what I thought I did already.

dbmcclain: fixed

src/Crypto/core-crypto.asd
dbmcclain: fixed

dbmcclain:
Thanks for taking a look at this stuff, Mark.

src/Crypto/crypto-pairings-test.asd
```diff
@@ -4,7 +4,7 @@
                core-crypto
                crypto-pairings)
   :perform  (test-op (o s)
-                     (symbol-call :lisp-unit :run-tests
+                     (symbol-call :lisp-unit :run-test
```
easye: @dbmcclain Actually, that needs to be `:run-tests` not `:run-test`, as thats the API of `lisp-unit` which actually runs the tests.

easye (changes requested):
Looking better.  

Now add `:core-crypto-test`  to the file `etc/all-tests.sexp`.  That way, the CI will run your tests as part of the test suite. 

It also means that 
```
(ql:quickload :asdf-test-harness)
(harness:run-suite :core-crypto-test)
``` 
will return a `nil` status if any of the tests in `:core-crypto-test` faile

dbmcclain:
okay we are significantly cleaned up and ready to merge into dev at your convenience

Trying to see what happened from the diff log will be a chore. What I did was to move all unused Crypto sources into a separate Emotiq-Unused folder. That is not being tracked, and I stopped the tracking on all those individual files.

If we ever need one of them again, I'll move it up into the Crypto folder again and restart tracking on it.

I also moved all our PBC-Interface, and ECC Curves, C-code into a Crypto-Libraries folder. They remain in separate folders because of significant differences in the makefile g++ command flags.

I did the same cleanup of old files for the Randhound folder.

And I fixed a few problems, sped up the Curve1174 modular inverse, and added test lines to the etc/all-tests file

PR #327, opened 2018-07-19
Refactored CI scripts, so they can be used locally

echupriyanov:


easye (commented):
A top-level README would be nice.

easye (commented):
I took a guess that I could just use `./install-{ccl,sbcl,lispworks}.bash` which I gave a spin under Linux on a fresh local Vagrant box.

CCL works.  

But for lispworks (which I don't expect to work completely because of licensing), I get the following error before licensing: 

```
usage: aws s3 cp <LocalPath> <S3Uri> or <S3Uri> <LocalPath> or <S3Uri> <S3Uri>
Error: Invalid argument type
Failure retrieving s3://emotiq-ci-supplements/lispworks/lw71-amd64-linux-patched.tar.gz.gpg
```

easye (approved):
Since both the CI and the local test harness invocation works, I approve from the standpoint of a developer.

PR #328, opened 2018-07-19
Added one line in BLOCK-FINISH handler to call for new regular electi…

dbmcclain:
…on. The Leader calls this unless a problem has arisen. In case of problems, system will timeout on emergency call for election and resync just like at genesis.

### DIFF
--- a/src/Cosi-BLS/cosi-handlers.lisp
+++ b/src/Cosi-BLS/cosi-handlers.lisp
@@ -179,7 +179,8 @@ THE SOFTWARE.
 (defmethod node-dispatcher ((msg-sym (eql :block-finished)) &key)
   (emotiq/tracker:track :block-finished)
   (emotiq:note "Block committed to blockchain")
-  (emotiq:note "Block signatures = ~D" (logcount (block-signature-bitmap (first *blockchain*)))))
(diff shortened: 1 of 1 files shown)


dbmcclain:
Looks like the CI pipeline is failing because, with recurrent elections, the tests never terminate...

easye:
> with recurrent elections, the tests never terminate...

Hmmm. Not sure of the right way to fix that.  

Can we add a parameter to `run-new-tx` to control whether elections are called again somehow?

Since we probably want the default to be to have elections, we need a way to turn elections off I am supposing.

dbmcclain:
one more special var I guess

let me know how you want to proceed. There are an accumulating number of special switches (vars) peppered over the source code already. At some point this will get out of hand. Needs a good organizing principle instead of ad-hoc, along with some foresight into setup and config

easye:
>There are an accumulating number of special switches (vars) peppered over the source code already. At some point this will get out of hand. Needs a good organizing principle instead of ad-hoc, along with some foresight into setup and config

Agreed. My naive first instinct would  be to centralize such things in `emotiq/config`

easye (changes requested):
Not merging until we figure how to keep the tests working.

PR #331, opened 2018-07-19
fix a nasty interaction between Actors and Emotiq/note

dbmcclain:
This uses a benign hook, stored as a property on the keyword symbol :ACTORS, to allow EMOTIQ:NOTE to override the behavior of Actors::blind-print.

If Actors is loaded first, it knows nothing about EMOTIQ package. And so it installs its own hook function = PRINT. But if Emotiq:NOTE is later loaded, it will override with a formatted call to NOTE.

If Actors is loaded after NOTE, then it leaves the hook alone, respecting the override.

If Actors is never loaded, no harm, no foul. Just a function stored on a property list of some symbol.

But it is important that all packages be able to view the hook without needing to know about any other package. Hence we use the keyword symbols package, which is ubiquitous.

### DIFF
--- a/src/Actors/actors-startup.lisp
+++ b/src/Actors/actors-startup.lisp
@@ -187,8 +187,13 @@ THE SOFTWARE.
 
 (defun blind-print (cmd &rest items)
   (declare (ignore cmd))
+  (let ((prfn (get :actors :print-handler)))
   (dolist (item items)
-    (emotiq:note "~A" item)))
+    (funcall prfn item))))
+
+(eval-when (:load-toplevel)
+  (unless (get :actors :print-handler)
+    (setf (get :actors :print-handler) 'print)))
 
 (defvar *shared-printer-actor*    #'blind-print)
 
@@ -201,8 +206,7 @@ THE SOFTWARE.
           (make-actor
            (dlambda
             (:print (&rest things-to-print)
-                    (dolist (item things-to-print)
-                      (emotiq:note "~A" item)))
+		    (apply 'blind-print :print things-to-print))
             
             (:quit () (become 'blind-print))
             )))
--- a/src/note.lisp
+++ b/src/note.lisp
@@ -41,6 +41,11 @@ a CL:FORMAT control string referencing the values contained in ARGS."
     (record-note timestring " " outstring)
     outstring))
 
+(eval-when (:load-toplevel)
+  ;; hook, even if Actors isn't loaded...
+  ;; If Actors are loaded later, they will respect the hook
+  (setf (get :actors :print-handler) (um:curry 'note "~A")))
+
 (defun em-warn (message-or-format &rest args)
   "Like note but this is for warnings."
   (let ((timestring (timestring))


src/Actors/actors-startup.lisp
```diff
@@ -187,8 +187,13 @@ THE SOFTWARE.
 
 (defun blind-print (cmd &rest items)
   (declare (ignore cmd))
+  (let ((prfn (get :actors :print-handler)))
   (dolist (item items)
-    (emotiq:note "~A" item)))
```
dbmcclain, line 191: if Actors is loaded first, it doesn't know about package EMOTIQ - kaboom!

src/Actors/actors-startup.lisp
```diff
@@ -201,8 +206,7 @@ THE SOFTWARE.
           (make-actor
            (dlambda
             (:print (&rest things-to-print)
-                    (dolist (item things-to-print)
```
dbmcclain, line 204: this was just redundant code. By vectoring to the "new" blind-print, we inherit the print hook too.

src/Actors/actors-startup.lisp
```diff
@@ -187,8 +187,13 @@ THE SOFTWARE.
...
 (defun blind-print (cmd &rest items)
   (declare (ignore cmd))
+  (let ((prfn (get :actors :print-handler)))
   (dolist (item items)
-    (emotiq:note "~A" item)))
+    (funcall prfn item))))
+
+(eval-when (:load-toplevel)
+  (unless (get :actors :print-handler)
+    (setf (get :actors :print-handler) 'print)))
```
dbmcclain, line 197: If Actors loads after EMOTIQ:NOTE, then leave the hook alone

src/note.lisp
```diff
@@ -41,6 +41,11 @@ a CL:FORMAT control string referencing the values contained in ARGS."
     (record-note timestring " " outstring)
     outstring))
 
+(eval-when (:load-toplevel)
+  ;; hook, even if Actors isn't loaded...
+  ;; If Actors are loaded later, they will respect the hook
+  (setf (get :actors :print-handler) (um:curry 'note "~A")))
+
```
dbmcclain, line 48: Adding an override hook on Actors, just in case Actors has been loaded and uses its PR function anywhere.

svspire:
Wow. I haven't used a symbol plist in forever, but this is a perfect use for it.
(Sorry I created this problem.)

dbmcclain:
No problem, glad that Lisp was so capable of making it easy to fix!!

:-)  DM

svspire:
`(setf (get :actors :print-handler) (um:curry 'note "~A"))`

needs to be

```
(setf (get :actors :print-handler)
      (lambda (&rest messages)
        (dolist (msg messages)
          (emotiq:note "~A" msg))))
```

The function resulting from `(um:curry 'note "~A")` will ignore second and later args.

svspire:
Never mind. I just saw that you're doing the dolist elsewhere.

ks (commented):
We could eventually merge this into DEV branch, as it's not directly related to emotiq/txn-package and that way also have UM package already in place?

svspire:
This branch has been subsumed into dev-us branch. Suggest we close this PR.

PR #332, opened 2018-07-19
Svs newtx

svspire:
Doesn't incorporate David's note hook (yet).


easye (changes requested):
Not approving until we can get the CI tests for the simulator fixed somehow.

svspire:
Subsumed by dev-us branch.

PR #333, opened 2018-07-19
Dbm recurrent elections fixed note hook

dbmcclain:
Ran into a problem under Mac CCL, where it does not have useful-macros loaded during quickload of emotiq/sim. So recoded the hook without needing anything but ubiquitous Lisp.


svspire:
This branch has been merged into dev-us branch. Suggest it be closed.

PR #358, opened 2018-07-21
Dev us

easye:
Potential merge from collected 20-JUL-2018 work


dbmcclain:
While there appear to be no conflicts, something seems to have changed while I was out to dinner. I had the system running for several hours on AWS. But now it bombs almost immediately because there are no entries in the witness pkey/stake list. 

Viewing code in Cosi-BLS/mvp-election-beacon.lisp shows an extension of code that I originally wrote, where it attempts on first call of the setup-emergency-call-for-new-election, that it should get the witness list from the genesis block. And apparently that list in the genesis block is now empty.

svspire:
The only thing that I know of that changed was that I reset the [rh-devnet] machines to use dev-us branch and restart with new keys and config files so that they know about each other. I didn't touch the other AWS clusters.

easye:
> This branch cannot be rebased due to conflicts

Sigh.

easye:
I didn't read Shannon's notes thoroughly enough.  This exactly duplicates  #340

PR #373, opened 2018-07-24
Run first node in CCL IDE for local blockchain

echupriyanov:
Instructions to run are in updated `README.md`

### DIFF
--- a/demo-network/README.md
+++ b/demo-network/README.md
@@ -95,6 +95,16 @@ One can attach to the node Lisp REPL using following command:
 tmux a -t node1   # for Node 1
 ```
 
+### (Experimental) Start blockchain with CCL IDE running 1st node
+```bash
+./start-blockchain-with-ide.bash
+```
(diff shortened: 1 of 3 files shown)


easye:
@svspire Can you please check that this satisfies your needs as you requested this, and that suitable code has been merged to `dev`?

svspire (changes requested):
Doesn't work.
open -a "Clozure CL64" $TMPFILE
doesn't work with the IDE; args passed this way get opened in the editor rather than loaded. I didn't realize this limitation when I asked for this capability. I will investigate other approaches. It might be possible to do it with Applescript.

echupriyanov:
@svspire Another option is to start IDE from the REPL. But that way listener will stay in `tmux`, but,  you will have IDE attached to the node process.

echupriyanov:
@svspire Addressing this in PR #403 . Please, take a look

PR #379, opened 2018-07-27
Dev us dbm self repairing

dbmcclain:
`*BLOCKCHAIN*` now contains the (integer) block hash of the latest block
used in conjunction with `*BLOCKCHAIN-TBL*`.

`(LATEST-BLOCK)` - returns the block corresponding to the ID in `*BLOCKCHAIN*`
`(BLOCK-LIST &OPTIONAL (FROM *BLOCKCHAIN*))` - returns a simple list, ordered from latest toward earliest, beginning at FROM. Blocks identified by integer hash code of the block header.

`(KILL-NODE PKEY)` - disable message dispatching to the indicated node (node identified by PKey)
`(ENABLE-NODE PKEY)` - enables message dispatching to the indicated node (if it was disabled)

Self repairing blockchain in each node. Leader broadcasts, to entire network, the ID of the latest block, even if no new block has been produced. If a node detects that its `*BLOCKCHAIN*` differs from the broadcast, it begins asking for backfill till it finds a block that it already knows about, or until genesis block (NIL prev-hash) is found. This backfilling occurs in a separate Actor so that the node remains responsive to normal message flow between nodes.

Backfilling occurs by asking a random node for the block. If that node has the block it is sent back to the requester, where it is validated and then added to its own blockchain. If the asked node does not have the block, it will ask one of its own neighbors (randomly selected) to try, as long as the depth of the forwarding is below `*MAX-QUERY-DEPTH*` (default to 8 levels). If requester doesn't get a valid answer within 60 sec, it tries a new query, up to 3 tries. If returned block points to another block that it doesn't know about, a new request is issued for that predecessor block. until it either already knows the predecessor block, or the genesis block has been reached. Back-filling occurs in parallel with normal message handling activities on a related but separate Actor. (related by referring to same NODE structure)

Block validation at the requestor is: (1) does the block header actually hash to the ID requested? (2) does the multisignature contain a BFT threshold number of signatories? (3) does the multisignature validate against the header hash? (4) do the block transactions actually hash to the header Merkel hash?

The :BLOCKCHAIN-HEAD message is broadcast by Leader separately from :HOLD-AN-ELECTION. The intent here is that :BLOCKCHAIN-HEAD Is to be widely broadcast to anyone listening, while :HOLD-AN-ELECTION is intended only for the staked witness pool. Eventually, there may become a distinction for selective broadcasts.  :BLOCKCHAIN-HEAD message is issued by Leader node for every epoch, even if no new block has been produced. So newly joining nodes need only wait for one election epoch to transpire before learning of blockchain head. And they automatically begin back filling by seeing that the broadcast blockchain head ID differs from their own, initially NIL, blockchain ID.

Clean up SHORT-ID to provide terser diagnostic displays

Explanatory comments added below...

### DIFF
--- a/src/Cosi-BLS/block.lisp
+++ b/src/Cosi-BLS/block.lisp
@@ -184,7 +184,7 @@ added to the blockchain."
       (when prev-block? 
         (with-slots (prev-block-hash)
             blk
-          (setf prev-block-hash (hash-block prev-block?))))
+          (setf prev-block-hash (int (hash-block prev-block?)))))
       blk)))
 
 
--- a/src/Cosi-BLS/block.lisp
+++ b/src/Cosi-BLS/block.lisp
@@ -176,7 +176,7 @@ added to the blockchain."
       (setf timestamp (- (get-universal-time) *unix-epoch-ut*))
       (setf election-proof block-election-proof)
       (setf leader-pkey block-leader-pkey)
-      (setf witnesses block-witnesses)
+      (setf witnesses (coerce block-witnesses 'vector))
       (setf transactions block-transactions)
       (setf merkle-root-hash (compute-merkle-root-hash blk))
       (setf input-script-merkle-root-hash (compute-input-script-merkle-root-hash blk))
--- a/src/Cosi-BLS/cosi-construction.lisp
+++ b/src/Cosi-BLS/cosi-construction.lisp
@@ -117,6 +117,38 @@ THE SOFTWARE.
              :initarg  :self)
    ))
 
+;; -------------------------------------------------------
+
+(defvar *current-node*  nil)  ;; for sim = current node running
+
+(defun current-node ()
+  *current-node*)
+
+(defmacro with-current-node (node &body body)
+  `(let ((*current-node* ,node))
+     ,@body))
+
+(define-symbol-macro *blockchain*     (node-blockchain     *current-node*))
+(define-symbol-macro *blockchain-tbl* (node-blockchain-tbl *current-node*))
+(define-symbol-macro *mempool*        (node-mempool        *current-node*))
+(define-symbol-macro *utxo-table*     (node-utxo-table     *current-node*))
+(define-symbol-macro *leader*         (node-current-leader *current-node*))
+(define-symbol-macro *tx-changes*     (node-tx-changes     *current-node*))
+(define-symbol-macro *had-work*       (node-had-work       *current-node*))
+
+;; election related items
+(define-symbol-macro *election-calls* (node-election-calls *current-node*))
+(define-symbol-macro *beacon*         (node-current-beacon *current-node*))
+(define-symbol-macro *local-epoch*    (node-local-epoch    *current-node*))
+
+(defun add-to-blockchain (node blk)
+  (with-current-node node
+    (let ((hashID (int (hash-block blk))))
+      (setf *blockchain* hashID
+            (gethash hashID *blockchain-tbl*) blk))))
+
+;; -------------------------------------------------------
+
 (defmethod node-bitmap ((node node))
   (ash 1 (node-bit node)))
 
@@ -150,10 +182,7 @@ THE SOFTWARE.
 
 (defmethod initialize-instance :after ((node node) &key &allow-other-keys)
   (when (null (node-blockchain node))
-    (let* ((genesis-block (emotiq/config:get-genesis-block))
-           (hashID        (hash-block genesis-block)))
-      (setf (node-blockchain node) hashID
-            (gethash hashID (node-blockchain-tbl node)) genesis-block))))
+    (add-to-blockchain node (emotiq/config:get-genesis-block))))
 
 ;; --------------------------------------------------------------
 
--- a/src/Cosi-BLS/cosi-construction.lisp
+++ b/src/Cosi-BLS/cosi-construction.lisp
@@ -195,9 +195,9 @@ THE SOFTWARE.
 (defun block-list (&optional (from *blockchain*))
   (um:accum acc
     (um:nlet-tail iter ((id from))
-      (um:when-let (blk (gethash id *blockchain-tbl*))
+      (um:when-let (blk (gethash (int id) *blockchain-tbl*))
         (acc blk)
-        (iter (int (block-prev-block-hash blk)))
+        (iter (block-prev-block-hash blk))
         ))))
 
 
--- a/src/Cosi-BLS/cosi-construction.lisp
+++ b/src/Cosi-BLS/cosi-construction.lisp
@@ -150,11 +150,10 @@ THE SOFTWARE.
 
 (defmethod initialize-instance :after ((node node) &key &allow-other-keys)
   (when (null (node-blockchain node))
-    (let ((genesis-block (emotiq/config:get-genesis-block)))
-      (push genesis-block (node-blockchain node))
-      (setf (gethash (cosi/proofs:hash-block genesis-block)
-                     (node-blockchain-tbl node))
-            genesis-block))))
+    (let* ((genesis-block (emotiq/config:get-genesis-block))
+           (hashID        (hash-block genesis-block)))
+      (setf (node-blockchain node) hashID
+            (gethash hashID (node-blockchain-tbl node)) genesis-block))))
 
 ;; --------------------------------------------------------------
 
--- a/src/Cosi-BLS/cosi-construction.lisp
+++ b/src/Cosi-BLS/cosi-construction.lisp
@@ -167,9 +167,9 @@ THE SOFTWARE.
   (and *blockchain*
        (gethash *blockchain* *blockchain-tbl*)))
 
-(defun block-list ()
+(defun block-list (&optional (from *blockchain*))
   (um:accum acc
-    (um:nlet-tail iter ((id *blockchain*))
+    (um:nlet-tail iter ((id from))
       (unless (null id)
         (let ((blk (gethash id *blockchain-tbl*)))
           (acc blk)
--- a/src/Cosi-BLS/cosi-construction.lisp
+++ b/src/Cosi-BLS/cosi-construction.lisp
@@ -195,10 +195,11 @@ THE SOFTWARE.
 (defun block-list (&optional (from *blockchain*))
   (um:accum acc
     (um:nlet-tail iter ((id from))
-      (um:when-let (blk (gethash (int id) *blockchain-tbl*))
-        (acc blk)
-        (iter (block-prev-block-hash blk))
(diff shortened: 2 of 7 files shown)


src/Cosi-BLS/cosi-construction.lisp
```diff
@@ -117,6 +117,38 @@ THE SOFTWARE.
              :initarg  :self)
    ))
 
+;; -------------------------------------------------------
```
dbmcclain, line 117: All pseudo-vars moved from cosi-handlers.lisp into cosi-construction.lisp

src/Cosi-BLS/block.lisp
```diff
@@ -176,7 +176,7 @@ added to the blockchain."
       (setf timestamp (- (get-universal-time) *unix-epoch-ut*))
```
dbmcclain, line 176: make block construction code responsible for internal content format, instead of caller

src/Cosi-BLS/cosi-construction.lisp
```diff
@@ -151,16 +182,27 @@ THE SOFTWARE.
```
dbmcclain, line 179: single function (ADD-TO-BLOCKCHAIN NODE BLK) now handles coordinated addition of new blocks to blockchain, keeping `*BLOCKCHAIN*` ID code and corresponding hashtable `*BLOCKCHAIN-TBL*` properly updated.

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -47,30 +47,6 @@ THE SOFTWARE.
   
 ;; -------------------------------------------------------
```
dbmcclain, line 48: moved to cost-construction.lisp

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -191,8 +166,30 @@ THE SOFTWARE.
 (defmethod node-dispatcher ((msg-sym (eql :block-finished)) &key)
```
dbmcclain, line 165: message handlers for self-repairing operations

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -201,6 +198,15 @@ THE SOFTWARE.
 
 ;; ------------------------------------------------------------------------------------
```
dbmcclain, line 198: add KILL-NODE and ENABLE-NODE to allow for testing of BFT dropouts, self repairing operations, etc.

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -209,20 +215,79 @@ THE SOFTWARE.
...
+              (when actor
+                (apply 'send actor ans))
+              ))
+           
+           (t (&rest msg)
+              (apply 'send beh msg))
+           ))))))
 
 ;; -------------------------------------------------------------------------------------
+;; Blockchain Sync
```
dbmcclain, line 234: Actual back-filling code, running in a separate but related Actor to keep the principal Actor responsive to normal message flow

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -647,10 +712,12 @@ check that each TXIN and TXOUT is mathematically sound."
   (um:accum acc
     (iter-subs node #'acc)))
 
-(defmethod short-id ((node node))
-  (short-id (node-pkey node)))
+;; ---------------------------------------------------------------------
```
dbmcclain, line 725: clean up SHORT-ID, and use (PR (MAPCAR 'SHORT-ID msg)) in main message dispatch for terser diagnostic output

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -886,7 +958,29 @@ check that each TXIN and TXOUT is mathematically sound."
 
 (defun done-with-duties ()
   (setup-emergency-call-for-new-election))
-  
+
+;; ----------------------------------------------------------------------
```
dbmcclain, line 973: Create / validate the `:BLOCKCHAIN-HEAD` message broadcast by Leader node, whether or not we add a new block this epoch

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -895,7 +989,7 @@ check that each TXIN and TXOUT is mathematically sound."
      (or
```
dbmcclain, line 1000: `*BLOCKCHAIN*` now contains only the hash ID (integer) of the most recent block. Get that most recent block with (LATEST-BLOCK)

src/Cosi-BLS/cosi-netw-xlat.lisp
```diff
@@ -91,14 +91,6 @@ THE SOFTWARE.
       (apply 'ac:send (node-self *my-node*) msg)
     (gossip-send pkey nil msg)))
         
-;; --------------------------------------------------------------
```
dbmcclain, line 94: Found this left-over unused code in cosi-netw-xlate.lisp, removed

src/Cosi-BLS/mvp-election-beacon.lisp
```diff
@@ -209,6 +209,14 @@ based on their relative stake"
    
 (defvar *mvp-election-period*  20) ;; seconds between election rounds
```
dbmcclain, line 211: NODE-STAKE slot in NODE structure no longer filled in because we get stakes from emotiq/config. So provide a way to find my own stake amount. Used only by diagnostic messaging.

src/Cosi-BLS/mvp-election-beacon.lisp
```diff
@@ -270,8 +277,7 @@ based on their relative stake"
 
 (defun setup-emergency-call-for-new-election ()
```
dbmcclain, line 276: NODE was unused here

src/Cosi-BLS/new-transactions.lisp
```diff
@@ -1056,7 +1056,7 @@ OBJECTS. Arg TYPE is implicitly quoted (not evaluated)."
    blocks of the blockchain (i.e., the value of cosi-simgen:*blockchain*)
```
dbmcclain, line 1056: get the list of blocks in the blockchain via (BLOCK-LIST)

src/Cosi-BLS/package.lisp
```diff
@@ -329,6 +329,10 @@
    :set-nodes
```
dbmcclain, line 329: Export new symbols in COSI-SIMGEN package

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -245,7 +245,6 @@ THE SOFTWARE.
   (let* ((bits  (block-signature-bitmap blk))
```
dbmcclain, line 246: tests already performed in multisignature validation function

src/Cosi-BLS/cosi-handlers.lisp
```diff
@@ -943,8 +942,10 @@ check that each TXIN and TXOUT is mathematically sound."
...
```
dbmcclain, line 956: add check that recorded Leader pkey is also present in witness list

easye:
Not all refactored references in existing code have been changed, as my simple "Run a chain, make a transaction" scenario fails with 

```
There is no applicable method for the generic function:
  #<STANDARD-GENERIC-FUNCTION COSI/PROOFS:BLOCK-PREV-BLOCK-HASH #x302002C1C47F>
when called with arguments:
````

To reproduce:  
```
(asdf:load-system :emotiq/startup)
(asdf:load-system :emotiq/txn)

(emotiq:main)

(let ((coin-account
       (let ((public-private (first (gossip/config:get-values))))
         (emotiq/config::make-keying-triple
          (first public-private)
          (second public-private)))))
  (emotiq/txn:get-utxos (emotiq/txn:address coin-account)))

```

**Review by easye — CHANGES_REQUESTED** [review]
Not all references in `cosi/proofs/newtx` seem to have been refactored.  

I did a quick check by running the payment scenario, and resulted in an error.  

Unfortunately we don't have systematic tests for `cosi/proofs/newtx`, so we need to be a little more careful before merging this change.

**Comment by dbmcclain** [comment]
Problem was that (INT NIL) = 0, not nil. So I put in check (UNLESS (ZEROP ID) ...)

**Comment by dbmcclain** [comment]
Final fix allows for termination when either ID = NIL (as from Genesis block), or upon a missing block in blockchain-tbl. 

I thought about triggering a back-fill operation here, but decided against that, as it is probably too general and blockchain might not even be running at this point.

**Comment by dbmcclain** [comment]
There is no direct way to "ask for a backfill". Thought about providing one, but it happens automatically when the blockchain is running and a new NODE joins the network. How you make that happen is beyond the purview of this primitive function.

**Comment by dbmcclain** [comment]
Add a user callable START-BACKFILL function. 

When you call (BLOCK-LIST) and receive a warning about a missing block, you can find the block ID of the missing block in the block header of the last block in the returned list. BLOCK-PREV-BLOCK-HASH slot of the block. You need that block ID and a reference to the NODE structure that needs the backfill.

When a new node joins the network, its `*BLOCKCHAIN*` will be nil, and so when the Leader of a running blockchain announces the head of the blockchain with a broadcast :BLOCKCHAIN-HEAD message, the new node will automatically call this function and start back filling for itself.

But if you run into a hole in your blockchain, you need to manually start the backfilling from that hole onward.

**Comment by easye** [comment]
@dbmcclain I still have problems with `cosi-simgen:block-list`. 

If I call it from the REPL, it fails, as the optional argument is nil:

```
There is no applicable method for the generic function:
  #<STANDARD-GENERIC-FUNCTION COSI-SIMGEN:NODE-BLOCKCHAIN #x302002C6A29F>
when called with arguments:
  (NIL)
```

If I explicitly "go hunting" for the last hash, 

```
(cosi-simgen:block-list (cosi-simgen:node-blockchain (first (gossip::local-real-nodes))))
```

the call still fails.  

A design goal here was to provide a simple function call that would return the local node's list of all blocks in the chain.

dbmcclain:
Heh! Yes, of course it will fail from the REPL. It needs a CURRENT-NODE, and you have none from the REPL. Remember that `*BLOCKCHAIN*` is a PSEUDO-Var. All pseudo-vars are referenced of off (CURRENT-NODE).

So, if you happen to know of a node, then try:
(WITH-CURRENT-NODE your-node (BLOCK-LIST))

dbmcclain:
You need to specify a NODE context via WITH-CURRENT-NODE before calling these functions. To support simulation, we need the context to be node-centric, not global.

- DM

PR #380, opened 2018-07-27
 Create top-level TRANSPORT networking module

lukego:
TRANSPORT is the new top-level name for the GOSSIP/TRANSPORT networking code. This code is actually independent of GOSSIP - they integrate via parameterizable callback functions - and so making it an independent module is simpler.

TRANSPORT/BLOCKING is the backend based on threads and blocking operations via the USOCKETS library.

TRANSPORT/ASYNC is the backend based on a non-blocking event loop with cl-async (libuv).

TRANSPORT/TEST contains some testing/benchmarking utility functions. These can be used interactively or combined with a test harness (not included.)

Includes @echupriyanov's CI support for libuv code cherry-picked from #306.


echupriyanov:
Original PR fails in CI with stack overflow in LispWorks. Fixed it with doubling stack size (16000 -> 32000)

echupriyanov (approved):
CI is passing now.
And we need clearly defined tests for new TRANSPORT system

svspire (changes requested):
Thanks for this Luke. We don't want to merge such a large change to the code base until we get the MVP working but we will try to test it soon.

lukego:
Thanks for the fixes @echupriyanov!

I have pushed unit tests but they are currently commented out. That's because `lisp-unit` is not installing via quicklisp on my dev machine. Likely a local problem.

One issue in the unit tests (mentioned in the commit msg) is that they assume that they can bind a specific TCP listen port. If this port is not available then the test will report a false-negative. This is unfortunate. I dropped the automatic port assignment feature because I didn't see how to support it with `cl-async`.

easye:
@lukego Could you redo the rebase against the latest `gossip` conflicts on `dev` when you get a chance?

easye (changes requested):
Could you redo the rebase against the latest gossip conflicts on dev when you get a chance?

PR #381, opened 2018-07-27
Self-setting NODE context

dbmcclain:
forcibly set *CURRENT-NODE* when new Node is constructed. For sanity sake. It will make node context equal to the last node ever constructed by default. But on a single NODE machine, as in AWS, it will allow for REPL interactions without needing to specify WITH-CURRENT-NODE. However, WITH-CURRENT-NODE will override context when necessary, as always before.

### DIFF
--- a/src/filesystem.lisp
+++ b/src/filesystem.lisp
@@ -73,3 +73,15 @@
     wallet))
 
 
+(defun new-temporary-directory (&key (root #p"/var/tmp/emotiq/"))
+  (loop
+     :with sub-directory = (make-pathname
+                            :defaults root
+                            :directory (append (pathname-directory root)
+                                               (list (symbol-name (gensym)))))
+     :until (not (probe-file sub-directory))
+     :finally (progn
+                (ensure-directories-exist sub-directory)
+                (return sub-directory))))
+       
+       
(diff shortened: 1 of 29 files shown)


easye (commented):
I think this is covered in <https://github.com/emotiq/emotiq/pull/389> but I need to double-check. 

@dbmcclain Can you please check that this is in `dev` now that #389 has been merged?

src/filesystem.lisp
```diff
@@ -73,3 +73,15 @@
     wallet))
 
 
+(defun new-temporary-directory (&key (root #p"/var/tmp/emotiq/"))
+  (loop
+     :with sub-directory = (make-pathname
```
easye: @dbmcclain Should be 

```
  :for sub-directory = (make-pathname
``` 

fixed in `dev` with <https://github.com/emotiq/emotiq/commit/ea35ebe2feab2edbf361c1e3438efc09b1726433>

dbmcclain: If you say so... I didn't write this, but if you like, I will edit here online

dbmcclain:
emotiq-config-test = system not found ?? 

Not found in dev branch either.

Searched hi and low over this branch and the dev branch, and I'll be darned if I can find any mention of emotiq-config-test. But it must be in there somewhere. Any hints?

dbmcclain:
At this point, these changes are almost 2 weeks old. I suggest strongly that you punt on them. This particular one was a simple 1-line addition to the initialize-instance method down in cosi-construction.lisp. That change now exists already in dev branch. So I think we should close this PR.

Issue #390, opened 2018-07-31
Inconsistent `:test-generate` behaviour


echupriyanov:
* running `env lisp=ccl systems=:test-generate etc/test-harness.bash` or `env lisp=lwpro systems=:test-generate etc/test-harness.bash` runs only first time, when there is no or empty `/var/tmp/emotiq` folder. Second run making Lisp process hang forever (I suspect on creating temporary folder(s), need investigation)
* sometimes, GENESIS-BLOCK assertions fail

echupriyanov:
@easye 
in function:
```lisp
(defun new-temporary-directory (&key (root #p"/var/tmp/emotiq/"))
  (loop
     :with sub-directory = (make-pathname
                            :defaults root
                            :directory (append (pathname-directory root)
                                               (list (symbol-name (gensym)))))
     :until (not (probe-file sub-directory))
     :finally (progn
                (ensure-directories-exist sub-directory)
                (return sub-directory))))
```
form `(symbol-name (gensym))` constantly returning the same value, resulting in infinite loop.
Proposed solution - use some form of random string instead.

easye:
@echupriyanov @dbmcclain @ks @vlebedev 

The failing test in #394 can be run on `dev` via:

```
(ql:quickload :test-generate)
(asdf:test-system :emotiq/config/generate)
```

The failing test is <https://github.com/emotiq/emotiq/blob/dev/src/test/genesis.lisp#L44> which may be invoked interactively in the REPL as 

```
(ql:quickload :test-generate)
(test-generate::create-genesis-block
```

echupriyanov:
Tests are different. in #394 I directly call `cosi/proofs:create-genesis-block` and check balance with generated block without serializing/deserializing block/keypair on disk.
@easye

easye:
@echupriyanov Good.  It confirms that the `lisp-encoder` serialization is not the culprit, which is a useful piece of information, but since that has already been working flawlessly (modulo the package and keyword issues we fixed in March), I thought it more likely that the keys were getting swapped.  But why does it only happen sometimes?  There are no actors/threads that I know of involved, so it doesn't make sense.

Maybe we should rename the original test to `mumble-with-serialization`?

echupriyanov:
I suggest moving to #394 to track the issue

easye:
Fix from David merged with <https://github.com/emotiq/emotiq/commit/6ecb31f1548c0ec4f63014457b273cab0c354bed>

PR #392, opened 2018-07-31
Fix infinite loop in `:emotiq/filesystem:new-temporary-directory`

echupriyanov:
Adresses #390


easye (approved):
Eugene submits his first patch to Lisp.  

NIce work.

Issue #394, opened 2018-08-01
Problems with Genesis Block


echupriyanov:
Trying to track down issue with `:test-generate` failures (#390), I've found, that sometimes `cosi/proofs:create-genesis-block` generates bad blocks.

I've created a simple test function:
```lisp
(defun create-and-check-genesis-block ()
  (let ((d (emotiq/filesystem:new-temporary-directory)))
    (let* ((nodes
            (emotiq/config/generate::generate-keys
             '((:hostname "127.0.0.1" :ip "127.0.0.1"))))
           (stakes
            (emotiq/config/generate::generate-stakes
             (mapcar (lambda (plist)
                       (getf plist :public))
                     nodes)))
           (address-for-coins
            (getf (first nodes) :public))
           (keypair (pbc:make-keying-triple (getf (first nodes) :public) (getf (first nodes) :private)))
           (genesis-block              
              (cosi/proofs:create-genesis-block address-for-coins stakes
              )))
        (values
         (cosi-simgen:with-block-list ((list genesis-block))
           (cosi/proofs/newtx:get-balance (emotiq/txn:address keypair)))
         d))))
```

I've added new test `VERIFY-GENESIS-BLOCK-GENERATE` to `:test-generate` system, running function above 100 times. At this moment it gives approx. 50/50 success/error rate.

dbmcclain:
Found and fixed the problem for you... see Twist https://twistapp.com/a/39362/ch/91845/t/438517/c/47411969

easye:
Fix from David merged in <https://github.com/emotiq/emotiq/commit/6ecb31f1548c0ec4f63014457b273cab0c354bed>.

PR #395, opened 2018-08-01
Added test to illustate issues with Genesis Block generation

echupriyanov:
Illustrates #394


easye (approved):
CI is failing; this adds a test of genesis block creation without serialization/deserialization through `lisp-object-encoder`.

Issue #385, opened 2018-07-30
Emotiq binary fails to start


echupriyanov:
Latest `emotiq` (built from `dev-eu` branch, fails to start:
```bash
eric@peccopa ~/C/E/e/v/t/v/l/p/emotiq-20180730011655-b996421b-macos> ./emotiq.bash 
workingdir = /Users/eric/Copoka/Emotiq/emotiq/var/tmp/var/local/production/emotiq-20180730011655-b996421b-macos
DYLD_LIBRARY_PATH /Users/eric/Copoka/Emotiq/emotiq/var/tmp/var/local/production/emotiq-20180730011655-b996421b-macos
DYLD_FALLBACK_LIBRARY_PATH /Users/eric/Copoka/Emotiq/emotiq/var/tmp/var/local/production/emotiq-20180730011655-b996421b-macos

Running from command line in production
with args [(/Users/eric/Copoka/Emotiq/emotiq/var/tmp/var/local/production/emotiq-20180730011655-b996421b-macos/emotiq)]

2018-07-30 15:10:54 Actor Directory created...

Running interactively in production
with args [(/Users/eric/Copoka/Emotiq/emotiq/var/tmp/var/local/production/emotiq-20180730011655-b996421b-macos/emotiq)]

2018-07-30 15:10:54 Not overwriting existing wallet named 'My Wallet'.
Error: Failed to make directory file /Users/vagrant/: Permission denied(13).
  1 (abort) Quit process.

Type :b for backtrace or :c <option number> to proceed.
Type :bug-form "<subject>" for a bug report template or :? for other options.

CL-USER 1 : 1 > :c 1
^C/Users/eric/Copoka/Emotiq/emotiq/var/tmp/var/local/production/emotiq-20180730011655-b996421b-macos/emotiq : Interrupted - quitting⏎                                                                                                   eric@peccopa ~/C/E/e/v/t/v/l/p/emotiq-20180730011655-b996421b-macos> 
```

easye:
@echupriyanov I have configuration changes that need to occur, namely unifying the filesystem for production binaries.

echupriyanov:
@easye fine. just placed it here, because it can block Misha's work on Wallet GUI

easye:
With <https://github.com/emotiq/emotiq/tree/mvp-websocket-mocks> I have hacked together a solution that will get further in starting the binary.  

Now, I am getting a problem with making PBC key pairs in my startup sequence.  
```
[...]

PR #396, opened 2018-08-01
init-pairing added to emotiq:start

guitarvydas:
This appears to fix issue #385 - pbc:init-pairings needs to be called during startup of binary


echupriyanov (approved):
Binary now starts ok

PR #398, opened 2018-08-01
Dev us sockets

svspire:
Detect and correct more error conditions in gossip-transport. No more half-dead dangling sockets.
Make pinger-daemon much more proactive, so that every node becomes connected to most or all other nodes much quicker.


dbmcclain:
I fetched branch dev-us-sockets -- how shall I test it?

dbmcclain (approved):
Shannon swears that if the CI agrees, we are good to go. So approved...

PR #400, opened 2018-08-02
Mvp websocket 20180802a

easye:
Package binary with configuration artifacts for websocket GUI mocks


echupriyanov (approved):
Built binary runs ok, and responds to WebSockets `ping` RPC call both on macOS and Linux

easye:
@guitarvydas This is not quite ready to merge to `dev`, as it creates a special set of configuration files that only runs the REST/websocket mocks.  

We need to get the `emotiq/app` package working, then generate artifacts for the single-process, multi-node simulation.

PR #402, opened 2018-08-02
fixed bug in select-inputs ; error was always being signalled

guitarvydas:
the error statement at the end of select-inputs was always signalled, moved it inside the cond and added more info on output

### DIFF
--- a/src/txn.lisp
+++ b/src/txn.lisp
@@ -77,8 +77,11 @@
          :do (incf selected-amount tx-amount)
          :do (push txn-spec selected-txns)
          :do (cond ((= selected-amount amount) (exact))
-                   ((> selected-amount amount) (surplus))))
-      (error "insufficient funds ~A" selected-txns))))
+                   ((> selected-amount amount) (surplus))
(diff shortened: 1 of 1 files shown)


dbmcclain (approved):
Looks like CI approves. And the fix looks correct. So I approve.

ks:
@guitarvydas @dbmcclain  

I'm slightly confused here with the fix, it seems to me that it introduces the error:

The cond here:

```
(cond ((= selected-amount amount) (exact))
          ((> selected-amount amount) (surplus))
```

is run for every UTXO until we either have exactly or too many coins from UTXOs - if that happens we return-from (via #'exact or #'surplus) - so the main function select-inputs succeeds...

It we iterate over all UTXOs and don't have enough, the loop terminates and then the error "insufficient funds" kicks in..

With this fix - if the first UTXO doesn't have enough coins, we take the T branch of the `cond` and error on insufficient funds.

ks:
Trouble is that so far (AFAIU) we don't have a test in CI which exercises a blockchain, which would catch that...

guitarvydas:
I agree.  I missed the return-froms in the flet.

PR #403, opened 2018-08-02
Run CCL IDE attached to tmux session

echupriyanov:
Launch node in REPL in `tmux` with CCL IDE attached.

@svspire please, take a look, if that suits you?
Run:
```bash
demo-network/start-node-with-ide.bash
```
To start node with IDE.

### DIFF
--- a/demo-network/start-node-with-ide.bash
+++ b/demo-network/start-node-with-ide.bash
@@ -31,13 +31,12 @@ node_name="node${node_id}"
 etc_and_wallets="${BASE}/node-configs/${node_name}/"
 
 cat >$TMPFILE <<EOF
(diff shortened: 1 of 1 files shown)


svspire (commented):
Just modified to use open command and --args parameter, which now works in modified version of CCL to pass args to the IDE.
Previous solution created a new, temporary IDE every time, which won't work well in practice.

echupriyanov:
@svspire Where I can find this modified version of CCL? I can document this in README and merge this PR into the `dev`. What do you think?

svspire:
It will be at https://github.com/svspire/ccl but I haven't pushed the changes yet. Still testing.

PR #404, opened 2018-08-02
changed hash= to int= in validate-cosi-message

guitarvydas:
on David's suggestion - this was failing in test-app and should probably have failed in anything that invokes the prepare phase


dbmcclain (approved):
I manually changed the second instance of HASH= to INT=. Now that CI approves, I do too.

PR #407, opened 2018-08-06
:PUBKEYS field in gossip-config should be LIST, not BIGNUM

echupriyanov:
Otherwise GOSSIP fails to start complaining about unknown public key


easye (approved):
I would rather have retained the distinction inside the `create-gossip-config` routine between whether gossip is creating a process with multiple gossip nodes or a single gossip node as this has ramifications of how the global hosts table will be constructed, but this is fine for now.

PR #410, opened 2018-08-06
Fix running Emotiq node in LWPro REPL.

### LINKED ISSUE #409: demo-network running with LispWorks doen't allow to use REPL

echupriyanov:
Instead of passing Lisp code through `-load` or stdin, run LWPro in `tmux` and send Lisp code through it to start the node.

Fixes #409


dbmcclain (approved):
sounds like a good idea!

PR #413, opened 2018-08-08
Fixing failing tests in :COSI-BLS-TEST and :TEST-GENERATE

### LINKED ISSUE #412: :COSI-BLS-TEST is failing for latest `dev-app-remove-overloading` branch
`public-key-to-key-hash-as-address-test` is failing because implicit conversion of strings/vectors to addresses no longer allowed.

echupriyanov:
Fixes #412 

* For :COSI-BLS-TEST we have to use explicit type conversion, no more accepting untagged strings and vectors
* For :TEST-GENERATE function `cosi/proofs/newtx:account-addresses=` was declaring itself as taking 2 strings as parameters, when now 2 `pbc:address` objects are passed (and compared actually)

### DIFF
--- a/src/Cosi-BLS/new-transactions.lisp
+++ b/src/Cosi-BLS/new-transactions.lisp
@@ -1384,7 +1384,7 @@ of type TYPE."
 ;;; this function.
 
 (defun account-addresses= (address-1 address-2)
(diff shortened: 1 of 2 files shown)


easye (approved):
Lookin' good.

PR #414, opened 2018-08-08
Dev app remove overloading

easye:
Cumulative work to no longer use references without type tags in our APIs.


echupriyanov (approved):
I think its good to go.
I suppose we are still missing updates to the Gossip part from Shannon, but that can be merged in separate PR

PR #418, opened 2018-08-08
Dev get transactions

guitarvydas:
This should add the ability to get transactions for a given account (including "change" transactions back to the account).

The big red section at the end is where I snipped out my transaction generator code (currently under test), to make this PR smaller and more manageable.


guitarvydas:
I (consciously) did another push - to remove some lines that I'd commented out earlier

dbmcclain (approved):
CI shows it working. I looked over your changes. But I don't know what problem you were trying to solve here. If it works for you, and since CI says okay, I will grant approval.

PR #428, opened 2018-08-14
Svs tagged pr

svspire:
Gossip uses only tagged objects as UIDs now. Gossip tests pass. Works in multinode networks when tested manually.


guitarvydas (commented):
This runs with test-app and produces the expected 6 blocks.  I'm set up only for running locally, so I have punted final approval to EC, DBM and MTE.  I, also tried merging DEV into this branch and had no problems.

echupriyanov (approved):
As far as it passes CI and smoke tests, I assume its fine to merge to dev and proceed from there.

PR #434, opened 2018-08-15
change how initial-coinbase-tx-in-id-value is created in new-transact…

### LINKED ISSUE #433: change how *initial-coinbase-tx-in-id-value* is created
form an arbitrary HASH/256 of any old seed, then copy the BEV into a new vector, then zap that vector with a FILL of zero. 
  And then reconstruct the HASH/256 item with that modified BEV. (COPY-SEQ)

guitarvydas:
Closes #433 
change how initial-coinbase-tx-in-id-value is created in new-transactions.lisp

### DIFF
--- a/src/Cosi-BLS/new-transactions.lisp
+++ b/src/Cosi-BLS/new-transactions.lisp
@@ -1000,9 +1000,20 @@ OBJECTS. Arg TYPE is implicitly quoted (not evaluated)."
 ;;; their normal function, so these values are arbitrary. These two ARE part
 ;;; of the hash of the transaction.
 
-(defvar *initial-coinbase-tx-in-id-value*
-  (make-instance 'hash:hash
-                 :val (bev (hex "0000000000000000000000000000000000000000000000000000000000000000"))))
+(defparameter *initial-coinbase-tx-in-id-value* nil)
+
+(defun zero-address ()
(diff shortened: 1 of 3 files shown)


src/Cosi-BLS/new-transactions.lisp
```diff
@@ -1000,9 +1000,20 @@ OBJECTS. Arg TYPE is implicitly quoted (not evaluated)."
...
+(defun zero-address ()
+  ; Form an arbitrary HASH/256 of any old seed, then copy the BEV into a new vector, then zap that vector with a FILL of zero. 
+  ; And then reconstruct the HASH/256 item with that modified BEV. (COPY-SEQ)
+  (let* ((h (hash:hash/256 :test))
+         (vec (vec-repr:bev-vec h))
+         (dum (copy-seq vec)))
+    (fill dum 0)
+    (setf (slot-value h 'hash::val) (vec-repr:bev dum))
+    h))
+
```
echupriyanov, line 1014: Maybe its better to define function `hash:zero/256` and use it?
Using `'hash::val` looks dangerous to me here.

guitarvydas: hash::val is the name of a slot, this code was developed by David
This whole PR is now superseded by #444

guitarvydas:
superseded by #444

PR #442, opened 2018-08-15
Fix initial txid3

### LINKED ISSUE #435: change EQUAL to PBC= in new-transactions.lisp

guitarvydas:
Closes #435 

In new-transactions.lisp, I found one EQUAL that I think should be PBC= and another that should simply be a numeric comparison (which I changed to =).


guitarvydas:
covered in new-transactions

